PR #31, opened 2013-12-23
First release of extra module m_geoipban

H7-25:
It add geoip country to whois and add extra ban  G: using iso country code

### DIFF
--- a/2.0/m_geoipban.cpp
+++ b/2.0/m_geoipban.cpp
@@ -1,0 +1,111 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2013 Filippo Cortigiani <simos@simosnap.org>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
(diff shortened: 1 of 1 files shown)


2.0/m_geoipban.cpp
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	void On005Numeric(std::string &output)
+	{
+		ServerInstance->AddExtBanChar('G');
+	}
+
+	virtual void OnWhois(User* src, User* dst)
```
Yusif-1: yeah, it adds their country to their whois? which make normal user can see anyone country and that not good 
most of user don't want any find that ,, it's their rights " privacy " matter
but it will be good if it's work only for ircops

NikosPapakonstantinou: That's not a problem since most users don't use full cloack.For example the last part of his hostname is visible: satmd (satmd@ChatSpikejt9qf1.de) Don't forget,you can use proxy to hide your country.Big networks don't have cloack and everyone can find your country easily.

Sheogorath-SI: Without the info in whois the whole module is useless for non-opers on a fullcloaded network.

The idea of the module is that a channel operator can ban a specific country from a channel. If you don't know where a users is from how can you ban him?

PR #56, opened 2014-10-15
Add m_requirectcp module which blocks clients not replying to ctcp.

NikosPapakonstantinou:
Any client without CTCP reply is blocked.

### DIFF
--- a/2.0/m_antibotctcp.cpp
+++ b/2.0/m_antibotctcp.cpp
@@ -16,7 +16,7 @@
  * along with this program.      If not, see <http://www.gnu.org/licenses/>.
  */
 #include "inspircd.h"
-/* $ModConfig: <antibotctcp ctcp="VERSION" msgonreply="true" accepted="Howdy buddy,you are authorized to use this server!"  declined="You have been blocked!Please get a better client."> */
+/* $ModConfig: <antibotctcp ctcp="VERSION" msgonreply="true" showlogs="true" accepted="Howdy buddy,you are authorized to use this server!"  declined="You have been blocked!Please get a better client."> */
 /* $ModDesc: Blocks clients not replying to CTCP like botnets/spambots/floodbots. */
 /* $ModAuthor: Nikos `UrL` Papakonstantinou */
 /* $ModAuthorMail: url@mirc.com.gr */
@@ -31,10 +31,10 @@ class ModuleAntiBotCTCP : public Module
 {
 	LocalIntExt ext;
 	bool msgonreply;
+	bool showlogs;
 	std::string accepted;
 	std::string declined;
 	std::string ctcp;
-	std::string tmp;
  public:
 	ModuleAntiBotCTCP()
 	: ext("ctcptime_wait", this)
@@ -58,6 +58,7 @@ class ModuleAntiBotCTCP : public Module
 	{
 		ConfigTag* tag = ServerInstance->Config->ConfValue("antibotctcp");
 		ctcp = tag->getString("ctcp");
+		showlogs = tag->getBool("showlogs", true);
 		msgonreply = tag->getBool("msgonreply", true);
 		declined = tag->getString("declined");
 		accepted = tag->getString("accepted");
@@ -82,8 +83,7 @@ class ModuleAntiBotCTCP : public Module
 	{
 		if (command == "NOTICE" && !validated && parameters.size() > 1 && ext.get(user))
 		{
-			tmp = "\001" + ctcp      + " ";
-			if (parameters[1].compare(0, tmp.length(), tmp) == 0)
+			if (parameters[1].size() > 1 && parameters[1][0] == 0x1 && parameters[1].compare(1, ctcp.length(), ctcp) == 0)
 			{
 				ext.set(user, 0);
 				if (msgonreply)
@@ -112,17 +112,18 @@ class ModuleAntiBotCTCP : public Module
 				if (declined.empty())
 				{
 					ServerInstance->Users->QuitUser(user, "You have been blocked!Please get a better client.");
-					ServerInstance->SNO->WriteGlobalSno('a', "Suspicious connection from %s (%s) was blocked by m_antibotctcp", user->GetFullRealHost().c_str(), user->GetIPString());
 				}
 				else
 				{
 					ServerInstance->Users->QuitUser(user, declined);
-					ServerInstance->SNO->WriteGlobalSno('a', "Suspicious connection from %s (%s) was blocked by m_antibotctcp", user->GetFullRealHost().c_str(), user->GetIPString());
 				}
 			}
 			else
 			{
 				ServerInstance->Users->QuitUser(user, "Disconnected");
+			}
+			if (showlogs)
+			{
 				ServerInstance->SNO->WriteGlobalSno('a', "Suspicious connection from %s (%s) was blocked by m_antibotctcp", user->GetFullRealHost().c_str(), user->GetIPString());
 			}
 			return MOD_RES_DENY;
--- a/2.0/m_antibotctcp.cpp
+++ b/2.0/m_antibotctcp.cpp
@@ -16,7 +16,7 @@
  * along with this program.      If not, see <http://www.gnu.org/licenses/>.
  */
 #include "inspircd.h"
-/* $ModConfig: <antibotctcp ctcp="VERSION" quitmsg="true" msgonreply="true" msg="If you are having problems connecting to this server, please get a better client."> */
+/* $ModConfig: <antibotctcp ctcp="VERSION" msgonreply="true" accepted="Howdy buddy,you are authorized to use this server!"  declined="You have been blocked!Please get a better client."> */
 /* $ModDesc: Blocks clients not replying to CTCP like botnets/spambots/floodbots. */
 /* $ModAuthor: Nikos `UrL` Papakonstantinou */
 /* $ModAuthorMail: url@mirc.com.gr */
@@ -29,10 +29,10 @@
 
 class ModuleAntiBotCTCP : public Module
 {
-	bool quitmsg;
-	bool msgonreply;
 	LocalIntExt ext;
-	std::string blockmsg;
+	bool msgonreply;
+	std::string accepted;
+	std::string declined;
 	std::string ctcp;
 	std::string tmp;
  public:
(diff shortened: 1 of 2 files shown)


Shawn-Smith:
Your coding style and indentation changes throughout the entire file. Lines 81-90 particularly are harder to read because of your indentation.

NikosPapakonstantinou:
Yeah true,these lines were looking good at my Notepad++.Something went wrong after the conversion.It's fixed now.

GladiaTeur:
Good Job Nikos.

BlackDragonGR:
Yeah, good job...

NikosPapakonstantinou:
It shows a quitmsg instead of [registration timeout] now.

darksis:
the best a module in inspircd

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.      If not, see <http://www.gnu.org/licenses/>.
+ */
+#include "inspircd.h"
+/* $ModConfig: <antibotctcp link="http://yoursite.gr" ctcp="VERSION" quitmsg="true" msgonreply="true"> */
+/* $ModDesc: Blocks clients not replying to CTCP like botnets/spambots/floodbots. */
+/* $ModAuthor: Nikos `UrL` Papakonstantinou */
+/* $ModAuthorMail: url@mirc.com.gr */
+/* $ModDepends: core 2.0-2.1 */
```
SaberUK: This should just be "core 2.0"

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,119 @@
...
+	void OnRehash(User* user)
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("antibotctcp");
+		quitmsg = tag->getBool("quitmsg", true);
+		msgonreply = tag->getBool("msgonreply", true);
+		ctcp = tag->getString("ctcp");
+		link = tag->getString("link");
+		if(ctcp.empty())
+		{
+			ServerInstance->SNO->WriteGlobalSno('a', "m_antibotctcp: Invalid ctcp value in config: %s", ctcp.c_str());
```
SaberUK: This should probably be done first to avoid changing all of the config when its broken.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	ModResult OnCheckReady(LocalUser* user)
+	{
+		if (ext.get(user))
+		{
+			if (quitmsg)
+			{
+				std::string qmsg;
+				qmsg = "If you are having problems connecting to this server, please get a better CLIENT or visit " + link + " for more info.";
```
SaberUK: You should make the message configurable rather than just the link.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,121 @@
...
+			blockmsg = "If you are having problems connecting to this server, please get a better client.";
+		}
+	}
+
+	ModResult OnUserRegister(LocalUser* user)
+	{
+		ConfigTag* tag = user->MyClass->config;
+		if (tag->getBool("antibotctcp", true))
+		{
+			user->WriteNumeric(931, "%s :Malicious or potentially unwanted softwares are not WELCOME here!", user->nick.c_str());
```
SaberUK: Why are you sending this numeric? Its only used in antibear because it cripples that specific bot.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser* user, bool validated, const std::string &original_line)
+	{
+		if (command == "NOTICE" && !validated && parameters.size() > 1 && ext.get(user))
+		{
+			if (parameters[1].compare(0, tmp.length(), tmp) == 0)
+			{
+				ext.set(user, 0);
+				if (msgonreply)
```
SaberUK: You could make this message configurable and then check if it is not `empty()` here.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	ModResult OnCheckReady(LocalUser* user)
+	{
+		if (ext.get(user))
+		{
+			if (quitmsg)
```
SaberUK: What exactly are you trying to do here? I don't think this does anything currently because OnCheckReady is called before OnUserRegister.

2.0/m_antibotctcp.cpp
NikosPapakonstantinou: Attila told me that i can sue 931 like antibear there but i can change it or remove it.

2.0/m_antibotctcp.cpp
NikosPapakonstantinou: Yeah but isn't it too much?Accepted Message / Blocked Message looks good yes.

2.0/m_antibotctcp.cpp
NikosPapakonstantinou: It does,it disconnects the user with a message because MOD_RES_DENY returns Registration Timeout instead.Handled by OnPreCommand otherwise it should block everyone.

2.0/m_antibotctcp.cpp
GLolol: Is such a notice is even needed - wouldn't a successful connect just be evident by the rest of the on-connect messages flowing in and regular commands starting to work?

2.0/m_antibotctcp.cpp
NikosPapakonstantinou: Plus numeric comes first there than the notice.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,133 @@
...
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser* user, bool validated, const std::string &original_line)
+	{
+		if (command == "NOTICE" && !validated && parameters.size() > 1 && ext.get(user))
+		{
+			tmp = "\001" + ctcp      + " ";
+			if (parameters[1].compare(0, tmp.length(), tmp) == 0)
```
SaberUK: You can get rid of tmp and do something like this:

```cpp
if (parameters.size() > 2 && parameters[1][0] == 0x1 && && parameters[1].compare(0, tmp.length(), tmp) == 0)
```

NikosPapakonstantinou: Oh thanks @SaberUK. Do you think ServerInstance->SNO->WriteGlobalSno L120 is okey?It floods the channel a bit but i don't know if it's usefull or not.

SaberUK: You might want to find an unused snomask character to use so if people find it too spammy they can turn it off.

NikosPapakonstantinou: Instead of finding an unused snomask,i added showlogs=true so they can enable/disable from the conf.

Cronus89:
Can this be merged now? Wanted to test it out, seems pretty neat.

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,134 @@
...
+		declined = tag->getString("declined");
+		accepted = tag->getString("accepted");
+	}
+
+	ModResult OnUserRegister(LocalUser* user)
+	{
+		ConfigTag* tag = user->MyClass->config;
+		if (tag->getBool("antibotctcp", true))
+		{
+			if (ctcp.empty())
```
SaberUK: You can get rid of this and block and pass "VERSION" as the second parameter to getString("ctcp").

2.0/m_antibotctcp.cpp
```diff
@@ -0,0 +1,134 @@
...
+	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser* user, bool validated, const std::string &original_line)
+	{
+		if (command == "NOTICE" && !validated && parameters.size() > 1 && ext.get(user))
+		{
+			if (parameters[1].size() > 1 && parameters[1][0] == 0x1 && parameters[1].compare(1, ctcp.length(), ctcp) == 0)
+			{
+				ext.set(user, 0);
+				if (msgonreply)
+				{
+					if (accepted.empty())
```
SaberUK: You can pass your default reason as a parameter to getString("accepted") and then do this:

```cpp
if (!accepted.empty()) 
  user->WriteServ(....);
````


Once you have done this you can get rid of msgonreply as people can just set an empty reason if they don't want a message.

File: 2.0/m_antibotctcp.cpp [hunk 10]
```diff
@@ -0,0 +1,134 @@
...
+
+	ModResult OnCheckReady(LocalUser* user)
+	{
+		if (ext.get(user))
+		{
+			if (msgonreply)
+			{
+				if (declined.empty())
+				{
+					ServerInstance->Users->QuitUser(user, "You have been blocked!Please get a better client.");
```
SaberUK: The same here applies as for what I said about accepted above.

SaberUK:
Minor suggestion: rename this module to something more accurate e.g. m_requirectcp.

Cronus89:
Nice, I like the changes. Module looks pretty good.  Anything else or can it be merged @SaberUK ?

2.0/m_requirectcp.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+	ModResult OnCheckReady(LocalUser* user)
+	{
+		if (ext.get(user))
+		{
+			if (!declined.empty())
+			{
+				ServerInstance->Users->QuitUser(user, declined);
+			}
+			ServerInstance->SNO->WriteGlobalSno('S', "Suspicious connection on port %d (class %s) from %s (%s) was blocked by m_requirectcp", user->GetServerPort(), user->MyClass->name.c_str(), user->GetFullRealHost().c_str(), user->GetIPString());
```
genius3000: WriteGlobalSno() should be given a lower case character. It will send globally to the uppercase automatically. That way 's' is local and 'S' is global.

2.0/m_requirectcp.cpp
```diff
@@ -0,0 +1,108 @@
...
+		ConfigTag* tag = ServerInstance->Config->ConfValue("requirectcp");
+		ctcp = tag->getString("ctcp", "VERSION");
+		declined = tag->getString("declined", "You have been blocked!Please get a better client.");
+		accepted = tag->getString("accepted", "Howdy buddy,you are authorized to use this server!");
+	}
+
+	ModResult OnUserRegister(LocalUser* user)
+	{
+		ConfigTag* tag = user->MyClass->config;
+		if (tag->getBool("requirectcp", true))
```
genius3000: Shouldn't this default to false to be 'opt-in' per connect class?

NikosPapakonstantinou: I think it's set to false by default. Add requirectcp=true inside the connect class to enable it otherwise it returns MOD_RES_PASSTHRU.

PR #67, opened 2015-06-25
Add a module for rotating log files.

SaberUK:
This module allows the automatic rotation of log files which use `strftime` specifiers in the file name.

For example,

     <log method="file" type="*" level="debug" target="%s.log">

### DIFF
--- a/2.0/m_rotatelog.cpp
+++ b/2.0/m_rotatelog.cpp
@@ -1,0 +1,77 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2015 Peter Powell <petpow@saberuk.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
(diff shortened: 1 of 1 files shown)


attilamolnar:
I think this should be named m_logrotate

SaberUK:
Does it really matter?

attilamolnar:
Yes, the system utility doing it is called logrotate so it makes sense to call the mod m_logrotate.

SaberUK:
It isn't at all related to the system utility though other than implementing something which is vaguely similar…

attilamolnar:
Thanks!

pgiblock:
This PR looks like the feature was merged, but I don't see any trace of it in inspircd:master (in past commits or in the current state of the branch.) Was this reverted for some reason? I've been waiting 3 years for this feature. Thanks

SaberUK:
@pgiblock You can install this module using `./modulemanager install m_rotatelog` from the root of the InspIRCd source directory.

PR #69, opened 2015-07-20
IRCv3.2 - Move cap-notify and invite-notify to main repo

AlphaT3ch:
Move cap-notify and invite-notify to main repo


SaberUK:
Yeah, no. These modules are extras for 2.0 and will be integrated in 2.2. They should not be removed from this repository.

PR #71, opened 2015-07-24
Module to command /IRCOPS that print out a list of current online and aways opers

antoniocostacf:
*fix to be used with protocol (raw 249)


attilamolnar:
Why not use or override /STATS p ?

antoniocostacf:
/STATS p, i found it at m_hideoper.cpp which means it's a rep of inspircd-2.0 not an extra.

PR #73, opened 2016-02-06
Add m_lusersnoservices and m_rpg modules

CyberBotX:
Upgraded my 1.2 versions to 2.0.

The m_lusersnoservices module adds a line to LUSERS that displays the global count without Services (assuming that Services are the only U-Lines on the network).

The m_rpg module adds the commands NPC, NPCA, AMBIANCE, NARRATOR and NARRATORA for use in pen and paper RPGs being played over IRC. For the 2.0 update, it has also been made to actually work on a network by broadcasting with ENCAP.

### DIFF
--- a/2.0/m_lusersnoservices.cpp
+++ b/2.0/m_lusersnoservices.cpp
@@ -1,0 +1,69 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2008-2016 Naram Qashat <cyberbotx@cyberbotx.com>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "inspircd.h"
+
+/* $ModDesc: Adds an extra line to LUSERS output to show global user count minus "clients" from U-Lines servers */
+/* $ModAuthor: Naram Qashat (CyberBotX) */
+/* $ModAuthorMail: cyberbotx@cyberbotx.com */
+/* $ModDepends: core 2.0 */
+
+class ModuleLusersNoServices : public Module
+{
+	int clientsNoServices;
+
+public:
+	void init()
+	{
+		Implementation implementations[] = { I_OnNumeric, I_OnPostConnect, I_OnUserQuit };
+		ServerInstance->Modules->Attach(implementations, this, sizeof(implementations) / sizeof(implementations[0]));
+
+		/* Calculate how many clients are not psuedo-clients introduced by the Services package */
+		this->clientsNoServices = 0;
+		user_hash::const_iterator curr = ServerInstance->Users->clientlist->begin(), end = ServerInstance->Users->clientlist->end();
+		for (; curr != end; ++curr)
+			if (!ServerInstance->ULine(curr->second->server) && !ServerInstance->ULine(curr->second->nick))
+				++this->clientsNoServices;
+	}
+
+	Version GetVersion()
+	{
+		return Version("Adds an extra line to LUSERS output to show global user count minus \"clients\" from U-Lines servers");
+	}
+
+	ModResult OnNumeric(User *user, unsigned numeric, const std::string &)
+	{
+		if (numeric == 266)
+			user->WriteNumeric(267, "%s :Current Global Users (Excluding Services): %d", user->nick.c_str(), this->clientsNoServices);
+		return MOD_RES_PASSTHRU;
+	}
+
+	void OnPostConnect(User *user)
+	{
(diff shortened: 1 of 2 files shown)


2.0/m_lusersnoservices.cpp
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	Version GetVersion()
+	{
+		return Version("Adds an extra line to LUSERS output to show global user count minus \"clients\" from U-Lines servers");
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser *user, bool validated, const std::string &original_line)
+	{
+		/* If the command doesnt appear to be valid, we dont want to mess with it. */
```
SaberUK: This is a very complex way of doing this. You would be better off hooking to OnWriteNumeric and when a numeric 266 is sent also sending your custom numeric.

2.0/m_lusersnoservices.cpp
```diff
@@ -0,0 +1,79 @@
...
+			{
+				std::vector<std::string> emptyParameters;
+				ServerInstance->CallCommandHandler(command, emptyParameters, user);
+			}
+
+			/* Calculate how many clients are not psuedo-clients introduced by the Services package */
+			int ClientsNoServices = 0;
+			user_hash::const_iterator curr = ServerInstance->Users->clientlist->begin(), end = ServerInstance->Users->clientlist->end();
+			for (; curr != end; ++curr)
+				if (!ServerInstance->ULine(curr->second->server) && !ServerInstance->ULine(curr->second->nick))
```
SaberUK: This is very performance intensive on large networks. You should calculate it once (on link?) and then cache it.

2.0/m_rpg.cpp
```diff
@@ -0,0 +1,374 @@
...
+		ServerInstance->Modules->Attach(I_OnPreCommand, this);
+	}
+
+	Version GetVersion()
+	{
+		return Version("Provides NPC, NPCA, AMBIANCE, NARRATOR, and NARRATORA commands for use by Game Masters doing pen & paper RPGs via IRC", VF_COMMON);
+	}
+
+	/** The purpose of this is to make it so the command text doesn't require a colon prefixing the text but also to allow a colon anywhere in the line.
+	 */
```
attilamolnar: Isn't it possible to do this inside the command handlers?

2.0/m_rpg.cpp
```diff
@@ -0,0 +1,374 @@
...
+/* $ModAuthor: Naram Qashat (CyberBotX) */
+/* $ModAuthorMail: cyberbotx@cyberbotx.com */
+/* $ModDepends: core 2.0 */
+
+#include "inspircd.h"
+
+/* Removes any ! characters from a given nick */
+static std::string strip_npc_nick(const std::string &nick)
+{
+	std::string newnick = "";
```
attilamolnar: `std::string newnick;` is enough to construct an empty string.

2.0/m_rpg.cpp
CyberBotX: As discussed on IRC, the reason for OnPreCommand being used as it is was to prevent a word starting with a colon from having its colon eaten by the command parser. This was mainly because I had people using the commands with smilies (like :D for instance). The issue is only a non-issue on broadcasting because I prefix a colon to the text before sending the ENCAP.

2.0/m_rpg.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+		/* Source is in the form of: *[nick]*!npc@[server-name] */
+		std::string npc_nick = strip_npc_nick(parameters[1]);
+		std::string npc_source = "*" + npc_nick + "*!npc@" + ServerInstance->Config->ServerName;
+
+		c->WriteChannelWithServ(npc_source, "PRIVMSG %s :%s%s%s", c->name.c_str(), action ? "\1ACTION " : "", this->text.c_str(), action ? "\1" : "");
+
+		if (localUser)
+		{
+			std::vector<std::string> params;
```
SaberUK: Can't this be done with GetRouting?

2.0/m_rpg.cpp
```diff
@@ -0,0 +1,374 @@
...
+			{
+				user->WriteNumeric(ERR_NOSUCHCHANNEL, "%s %s :No such channel", user->nick.c_str(), parameters[0].c_str());
+				return CMD_FAILURE;
+			}
+		}
+		else
+			this->text = parameters[1];
+
+		/* Source is in the form of: -Narrator-!npc@[server-name] */
+		std::string narr_source = std::string("-Narrator-!npc@") + ServerInstance->Config->ServerName;
```
SaberUK: This is a valid nickname. Maybe you should reserve it?

2.0/m_rpg.cpp
CyberBotX: No, because of a similar issue to why I had to implement the OnPreCommand override in the module itself. If the user types in some text and one of the words starts with a colon, it gets eaten on the broadcast.

SaberUK: If the users text begins with : then they should use two : to stop that.

attilamolnar: @SaberUK The idea is to not make the users do that

2.0/m_rpg.cpp
CyberBotX: How is -Narrator- a valid nickname? I cannot set that as my nick because it starts with a dash.

SaberUK: You can't with the default but you can with m_nationalchars.

attilamolnar:
Please squash the commits in a way that there are no bugfix commits.

CyberBotX:
Commits have been squashed into one commit.

attilamolnar:
Thanks!

PR #74, opened 2016-02-17
Fix m_rpg to split a message into multiple lines...

CyberBotX:
… if the given input would exceed the 512 character limit.

The reason for the loop is more for the just-in-case situations of the source and channel also being very long. I think the chances of getting more than 2 messages from a single usage of one of m_rpg's commands is highly unlikely, but this code should cove it if it somehow happens.

### DIFF
--- a/2.0/m_rpg.cpp
+++ b/2.0/m_rpg.cpp
@@ -42,6 +42,40 @@ static std::string strip_npc_nick(const std::string &nick)
 	return newnick;
 }
 
+/* Sends a message to a channel, splitting it as needed if the message is too long.
+ * It'll usually only split into 2 messages because of the 512 character limit. */
+static void send_message(Channel *c, const std::string &source, std::string text, bool action)
+{
+	/* 510 - colon prefixing source - PRIVMSG - colon prefixing text - 3 spaces = 498
+	 * Subtracting the source and the channel name to get how many characters we are allowed left
+	 * If doing an action, subtract an additional 9 for the startind and ending ASCII character 1, ACTION and space */
+	int allowedMessageLength = 498 - source.size() - c->name.size() - (action ? 9 : 0);
+	/* This will keep attempting to determine if there is text to send */
+	do
+	{
+		std::string textToSend = text;
+		/* Check if the current text to send exceeds the length allowed */
+		if (textToSend.size() > allowedMessageLength)
+		{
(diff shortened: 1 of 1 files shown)


2.0/m_rpg.cpp
```diff
@@ -42,6 +42,41 @@ static std::string strip_npc_nick(const std::string &nick)
...
 
+/* Sends a message to a channel, splitting it as needed if the message is too long.
+ * It'll usually only split into 2 messages because of the 512 character limit. */
+static void send_message(Channel *c, const std::string &source, const std::string &origText, bool action)
+{
+	/* 510 - colon prefixing source - PRIVMSG - colon prefixing text - 3 spaces = 498
+	 * Subtracting the source and the channel name to get how many characters we are allowed left
+	 * If doing an action, subtract an additional 9 for the startind and ending ASCII character 1, ACTION and space */
+	int allowedMessageLength = 498 - source.size() - c->name.size() - (action ? 9 : 0);
+	std::string text = origText;
```
attilamolnar: Pass it by value if you are going to copy it anyway

2.0/m_rpg.cpp
```diff
@@ -42,6 +42,41 @@ static std::string strip_npc_nick(const std::string &nick)
...
+			size_t lastSpace = textToSend.find_last_of(' ', allowedMessageLength);
+			/* If a space was found, split off the text to send */
+			if (lastSpace != std::string::npos)
+			{
+				textToSend = text.substr(0, lastSpace);
+				text = text.substr(lastSpace + 1);
+			}
+			/* Otherwise, we'll send whatever we have left, even if it may be too long */
+			else
+				text = "";
```
attilamolnar: `text.clear();`

CyberBotX:
I applied those changes and replaced the commit with a new one.

attilamolnar:
Okay, I haven't had the chance to test your work nor review it in-depth, but I'll merge it like this if that's fine by you.

CyberBotX:
I'm fine with that. I did a test on my own server of sending a command that was very much over 512 characters in length, and while the command as processed by InspIRCd to truncate it to 512 characters, the output would've been longer, so it split the message into two separate messages.

attilamolnar:
Merged, thanks!

PR #75, opened 2016-03-04
New apacheauth module

vccu:
Authenticates users against an Apache auth file.

### DIFF
--- a/2.0/m_apacheauth.cpp
+++ b/2.0/m_apacheauth.cpp
@@ -1,0 +1,162 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2011-2016 Collabora Ltd <vincent@collabora.co.uk>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <fstream>
+#include "inspircd.h"
+#include "hash.h"
+
+/* $ModDesc: Allow/Deny connections based upon an apache auth file */
+/* $LinkerFlags: -lcrypt */
+
+class ModuleApacheAuth : public Module
+{
+	std::string authfile;
+	std::string killreason;
+	std::string allowpattern;
+	bool verbose;
+	struct HashedPassword {
+		std::string algorithm;
+		std::string salt;
+		std::string hash;
+		std::string hashed;
+	};
+	std::map<std::string, HashedPassword> logins;
+
+ public:
+	ModuleApacheAuth()
+	{
+	}
+
+	void init()
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnRehash, I_OnUserRegister };
+		ServerInstance->Modules->Attach(eventlist, this, 2);
+	}
+
+	void OnRehash(User* user)
+	{
+		ConfigTag* conf = ServerInstance->Config->ConfValue("apacheauth");
+		authfile = conf->getString("authfile");
+		killreason = conf->getString("killreason");
+		allowpattern = conf->getString("allowpattern");
+		verbose = conf->getBool("verbose");
+		LoadAuthFile();
+	}
+
+	bool SplitRecord(const std::string &record, std::string &record_algorithm, std::string &record_salt, std::string &record_hash)
+	{
+		size_t i[3];
+
+		for (size_t pos = 0, n = 0; n < 3; ++n) {
+			i[n] = pos = record.find('$', pos);
+			if (pos == std::string::npos) return false;
+			++pos;
+		}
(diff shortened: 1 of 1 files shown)


2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	bool SplitRecord(const std::string &record, std::string &record_algorithm, std::string &record_salt, std::string &record_hash)
+	{
+		const char *s = record.c_str();
+		const char *ptr;
+
+		if (record[0]!='$') return false;
+		s = s+1;
+		ptr = strchr(s, '$');
```
SaberUK: You can use std::string::find here instead of using `const char*` and `strchr` here.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,168 @@
...
+		return true;
+	}
+
+	void LoadAuthFile()
+	{
+		logins.clear();
+		char s[2048];
+		int line=0;
+		ServerInstance->SNO->WriteGlobalSno('a', "Loading auth file %s",authfile.c_str());
+		FILE *f = fopen(authfile.c_str(), "r");
```
SaberUK: Using std::ifstream might make this code cleaner.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,168 @@
...
+		std::map<std::string,std::string>::const_iterator i = logins.find(user->ident);
+		if (i == logins.end()) {
+			ServerInstance->SNO->WriteGlobalSno('a', "Forbiding connection from %s!%s@%s (login not found)",
+				user->nick.c_str(), user->ident.c_str(), user->host.c_str());
+			ServerInstance->Users->QuitUser(user, killreason);
+			return MOD_RES_PASSTHRU;
+		}
+
+		std::string record_algorithm, record_salt, record_hash;
+		if (!SplitRecord(i->second, record_algorithm, record_salt, record_hash)) {
```
SaberUK: You should probably do this once on load and keep it in a custom structure.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,168 @@
...
+			return MOD_RES_PASSTHRU;
+		}
+		ServerInstance->SNO->WriteGlobalSno('a', "Granting access to connection from %s!%s@%s",
+				user->nick.c_str(), user->ident.c_str(), user->host.c_str());
+		return MOD_RES_PASSTHRU;
+	}
+
+	Version GetVersion()
+	{
+		return Version("Allow/Deny connections based upon an Apache auth file", VF_VENDOR);
```
SaberUK: This is not a vendor module. Don't use VF_VENDOR.

vccu:
All done. I pushed -f to the same branch, not sure this is what I'm meant to do, or push a new branch instead. Let me know if I should do the second way.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,163 @@
...
+		std::ifstream f;
+
+		logins.clear();
+		ServerInstance->SNO->WriteGlobalSno('a', "Loading auth file %s",authfile.c_str());
+		f.open(authfile.c_str(), std::ifstream::in);
+		if (!f.good()) {
+			ServerInstance->SNO->WriteGlobalSno('a', "Auth file failed to open, no connections will be allowed");
+			return;
+		}
+		while (1) {
```
SaberUK: You can do something like this and avoid using a C buffer and an infinite loop.

```c++
std::string line;
while (std::getline(stream, line))
{
    // do something with line here.
}

**Comment by vccu** [comment]
Done, pushed -f again.

File: 2.0/m_apacheauth.cpp [hunk 6]
```diff
@@ -0,0 +1,161 @@
...
+		logins.clear();
+		ServerInstance->SNO->WriteGlobalSno('a', "Loading auth file %s",authfile.c_str());
+		f.open(authfile.c_str(), std::ifstream::in);
+		if (!f.good()) {
+			ServerInstance->SNO->WriteGlobalSno('a', "Auth file failed to open, no connections will be allowed");
+			return;
+		}
+		std::string s;
+		while (std::getline(f, s)) {
+			if (!f.good()) break;
```
SaberUK: I think this is unnecessary now.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,161 @@
...
+			++line;
+			if (!s[0] || s[0] == '\r' || s[0] == '\n') continue;
```
SaberUK: Same for this.

vccu:
Changed and pushed -f again.

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,159 @@
+/*       +------------------------------------+
```
rburchell: this header style changed a while ago. it'd be nice if you could use the newer style (and your own copyright info)

2.0/m_apacheauth.cpp
```diff
@@ -0,0 +1,159 @@
...
+		if (!tag->getBool("useapacheauth", true))
+			return MOD_RES_PASSTHRU;
+
+		if (!allowpattern.empty() && InspIRCd::Match(user->ident,allowpattern))
+			return MOD_RES_PASSTHRU;
+
+		//ServerInstance->SNO->WriteGlobalSno('a', "Got request to auth %s with hash %s", user->nick.c_str(), user->password.c_str());
+		std::map<std::string, HashedPassword>::const_iterator i = logins.find(user->ident);
+		if (i == logins.end()) {
+			ServerInstance->SNO->WriteGlobalSno('a', "Forbiding connection from %s!%s@%s (login not found)",
```
rburchell: typo in message, I'd suggest "Denying connection" instead perhaps.

while you're here, can I suggest that you remove the commented out lines?

vccu:
Done.

PR #77, opened 2016-04-28
m_tgchange

Adam-:


2.0/m_tgchange.cpp
```diff
@@ -0,0 +1,247 @@
...
+	{
+		Implementation eventlist[] = { I_OnRehash, I_OnUserPreMessage, I_OnUserPreNotice, I_OnUserPreInvite };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist) / sizeof(Implementation));
+
+		OnRehash(NULL);
+	}
+
+	Version GetVersion()
+	{
+		return Version("Implements \"target change\" detection to help prevent spam bots", VF_NONE);
```
SaberUK: VF_VENDOR.

2.0/m_tgchange.cpp
```diff
@@ -0,0 +1,247 @@
...
+#include "inspircd.h"
+
+enum
+{
+	ERR_TARGCHANGE = 707
+};
+
+namespace
+{
+	unsigned TGCHANGE_NUM;   /* how many targets we keep track of */
```
SaberUK: Probably shouldn't be in screaming snake case?

2.0/m_tgchange.cpp
```diff
@@ -0,0 +1,247 @@
...
+			Channel *c = *i;
+
+			if (c->GetPrefixValue(source) >= VOICE_VALUE && std::find(target->chans.begin(), target->chans.end(), c) != target->chans.end())
+				return true;
+		}
+
+		return false;
+	}
+
+	ModResult Target(User *source, User *dest)
```
SaberUK: *& should be against the type not the variable name.

2.0/m_tgchange.cpp
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+ private:
+	bool Allowed(LocalUser *source, User *target)
+	{
+		for (UCListIter i = source->chans.begin(); i != source->chans.end(); i++)
+		{
+			Channel *c = *i;
+
+			if (c->GetPrefixValue(source) >= VOICE_VALUE && std::find(target->chans.begin(), target->chans.end(), c) != target->chans.end())
```
attilamolnar: User::chans is a set in 2.0; use `count()`

PR #78, opened 2016-04-29
m_con_throttle

Adam-:


2.0/m_con_throttle.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+		Throttle*& throttle = throttles[client->addr()];
+		if (throttle != NULL && ServerInstance->Time() - throttle->last_attempt < throttle_time)
+		{
+			if (throttle->con_count >= throttle_num)
+			{
+				if (sock->bind_tag->getString("ssl").empty())
+				{
+					const char err[] = "ERROR :Trying to reconnect too fast.\r\n";
+					send(fd, err, sizeof(err) - 1, 0);
```
SaberUK: You should use `ServerInstance->SE->Send` here.

Adam-: There is no eventhandler object here

2.0/m_con_throttle.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+/* $ModAuthor: Adam */
+/* $ModAuthorMail: Adam@anope.org */
+/* $ModDesc: Incoming connection throttle */
+/* $ModDepends: core 2.0 */
+
+class Throttle
+{
+ public:
+	int con_count;
```
attilamolnar: Should be an `unsigned int`

2.0/m_con_throttle.cpp
```diff
@@ -0,0 +1,127 @@
...
+};
+
+class ModuleConnThrottle : public Module
+{
+	typedef nspace::hash_map<std::string, Throttle*> throttle_hash;
+
+	throttle_hash throttles;
+
+	int throttle_num;
+	int throttle_time;
```
attilamolnar: Unsigned

2.0/m_con_throttle.cpp
```diff
@@ -0,0 +1,127 @@
...
+ public:
+	int con_count;
+	time_t last_attempt;
+
+	Throttle() : con_count(0), last_attempt(0) { }
+};
+
+class ModuleConnThrottle : public Module
+{
+	typedef nspace::hash_map<std::string, Throttle*> throttle_hash;
```
attilamolnar: Why are Throttle objects allocated separately?

attilamolnar:
Module name should begin with `m_conn` because of the existing m_conn_join, m_conn_umodes and m_conn_waitpong modules being named that way.

SaberUK:
Renamed and committed in 49387ff.

PR #83, opened 2016-07-09
Fix syntax notice in m_require_auth

MuffinMedic:
As with g-line, syntax should be in `[user@host]`, not `<nick>`

### DIFF
--- a/2.0/m_require_auth.cpp
+++ b/2.0/m_require_auth.cpp
@@ -144,7 +144,7 @@ class CommandGALine: public Command
 		: Command(c, linetype+"LINE", 1, 3)
 	{
 		flags_needed = 'o';
-		this->syntax = "<nick> [<duration> :<reason>]";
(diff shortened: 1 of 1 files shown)


SaberUK:
Why change to square brackets? They imply the parameter is optional when it is not.

MuffinMedic:
@SaberUK Kept the same syntax as `/gline` for consistency. No objections to it being changed but it should be changed across all similar commands. gline/zline/eline/etc... Currently they all have brackets around the user identifier.

SaberUK:
GLINE uses 

```cpp
syntax = "<ident@host> [<duration> :<reason>]";
```

MuffinMedic:
@SaberUK Sorry about that, I was looking at `helpop`, not the actual `/gline` output. I've update the pull request accordingly.

attilamolnar:
Please squash this into a single commit

attilamolnar:
Thanks, merged.

PR #64, opened 2015-05-15
Backport m_opmoderated.cpp (cmode +U) from 2.1

GLolol:
Tested working. I imported the changes from my [own repository](https://github.com/OVERdrive-IRC/inspircd-modules/commits/master) as a squash merge, which includes the following:
- OVERdrive-IRC/inspircd-modules@c624c09 Change extban from `U:` to `u:` to prevent conflicts with m_services_account
- OVERdrive-IRC/inspircd-modules@1527108 Quick fixes so that the module compiles on InspIRCd 2.0

PR #87, opened 2016-09-06
Added m_joinpartspam: chmode to ban joinpart spammers.

genius3000:
This adds channel mode +U with a 'cycles:sec' parameter. A cycle is counted as a successful join to the channel. Upon reaching the limit, the next attempted join (prior to the time reset) results in a WildHost ban and denying the join.

I'm using a mask of 'user@host' instead of 'User*' for tracking so a user is still caught after a quit, connect, and join. Not using the nickname in the mask so it's not so easily bypassed.

### DIFF
--- a/2.0/m_joinpartspam.cpp
+++ b/2.0/m_joinpartspam.cpp
@@ -1,0 +1,243 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2016 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "inspircd.h"
+
+/* $ModDesc: Adds channel mode +V to ban a user after x per y joins and parts/quits (join/part spam) */
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDepends: core 2.0 */
+
+/* Special thanks to Attila for the patience and guidance in
+ * fixing up some of my initial, poor methods.
+ */
+
+class joinpartspamsettings
+{
+ public:
+	unsigned int cycles;
+	unsigned int secs;
+	time_t lastcleanup;
+
+	struct Tracking
+	{
+		time_t reset;
+		unsigned int counter;
+		Tracking() : reset(0), counter(0) { }
+	};
+	std::map<std::string, Tracking> cycler;
+
+	joinpartspamsettings(unsigned int c, unsigned int s)
+		: cycles(c), secs(s), lastcleanup(ServerInstance->Time()) { }
+
+	/* Called by PostJoin to possibly reset a cycler's Tracking and increment the counter */
+	void addcycle(const std::string& mask)
+	{
+		/* If mask isn't already tracked, set reset time
+		 * If tracked and reset time is up, reset counter and reset time
+		 * Also assume another server denied join and set ban, with a user removing
+		 * the ban if counter >= cycles, reset counter and reset time
+		 */
+		Tracking& tracking = cycler[mask];
+
+		if (tracking.reset == 0)
+			tracking.reset = ServerInstance->Time() + secs;
+		else if (ServerInstance->Time() > tracking.reset || tracking.counter >= cycles)
+		{
+			tracking.counter = 0;
+			tracking.reset = ServerInstance->Time() + secs;
+		}
+
+		++tracking.counter;
+
+		this->cleanup();
+	}
+
+	/* Called by PreJoin to check if a cycler's counter exceeds the set cycles,
+	 * Will first clear a cycler if their reset time is up
+	 */
+	bool zapme(const std::string& mask)
+	{
+		/* Only check reset time and counter if they are already tracked as a cycler */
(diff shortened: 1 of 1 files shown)


GLolol:
Cmode +U conflicts with #64 :(

Techman-:
Network ops still have the ability to pick and choose what modules to use :)

Configurable modechar would be good.

genius3000:
I just had a look through helpop and went with an available letter. Could try find another one.
It is easy to change, just one spot in the code (other than comments).

genius3000:
Updated this with a few code style fixes and changed to chmode V. I know that could conflict with customprefix->Half voice, but I've never seen that used.

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,226 @@
...
+	void cleanup()
+	{
+		/* 10 minutes should be a reasonable wait time */
+		if (ServerInstance->Time() - lastcleanup < 600)
+			return;
+
+		for (std::map<std::string, time_t>::iterator i = resets.begin(); i != resets.end(); i++)
+		{
+			if (ServerInstance->Time() > resets[i->first])
+				this->clear(i->first);
```
attilamolnar: This call invalidates the iterator, if the entry needs to be removed ++ the iterator before calling the function

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,231 @@
...
+	time_t lastcleanup;
+
+	struct Tracking
+	{
+		time_t reset;
+		unsigned int counter;
+	};
+	std::map<std::string, Tracking> cycler;
+
+	joinpartspamsettings(int c, int s) : cycles(c), secs(s)
```
attilamolnar: These parameters should be `unsigned int`.

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,231 @@
...
+		 */
+		if (cycler.count(mask) != 1)
+			cycler[mask].reset = ServerInstance->Time() + secs;
+		else if (ServerInstance->Time() > cycler[mask].reset || cycler[mask].counter >= cycles)
+		{
+			cycler[mask].counter = 0;
+			cycler[mask].reset = ServerInstance->Time() + secs;
+		}
+
+		++cycler[mask].counter;
```
attilamolnar: Each `cycler[mask]` causes a map lookup which is wasteful because you are only interacting with the same element throughout the method. To improve this, do something like:
`Tracking& tracking = cycler[mask];`
then replace `cycler[mask]` with `tracking` in the rest of the method.

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,231 @@
...
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDepends: core 2.0 */
+
+class joinpartspamsettings
+{
+ public:
+	unsigned int cycles;
+	unsigned int secs;
+	time_t lastcleanup;
```
attilamolnar: This must be given a value before line 88 checks it. As I understand the initial value has to pass the check on that line for the code to work as intended.

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,231 @@
...
+		{
+			if (ServerInstance->Time() > cycler[mask].reset)
+				cycler.erase(mask);
+			else if (cycler[mask].counter >= cycles)
+			{
+				cycler.erase(mask);
+				return true;
+			}
+		}
+		return false;
```
attilamolnar: What you can do here to avoid repeated map lookups is to use the `find()` map method like this:

```c++
std::map<std::string, Tracking>::iterator it = cycler.find(mask);
if (it == cycler.end()) // 'mask' is not in the map
    return false;
Tracking& tracking = it->second;

// Rest of the code
if (ServerInstance->Time() > tracking.reset)
 ...
```

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	/* Clear expired entries of non cyclers */
+	void cleanup()
+	{
+		/* 10 minutes should be a reasonable wait time */
+		if (ServerInstance->Time() - lastcleanup < 600)
+			return;
+
+		for (std::map<std::string, Tracking>::iterator i = cycler.begin(); i != cycler.end(); )
```
attilamolnar: The same thing applies for `cycler[mask]` and `erase(key)` as above, they must find the element in the map but you already know where it is.
There's existing code showing how to do this more efficiently: https://github.com/inspircd/inspircd/blob/master/src/coremods/core_dns.cpp#L658-L665
That way, no extra strings get created and extra map lookups don't occur.

genius3000:
@attilamolnar how's this looking now?

attilamolnar:
@genius3000 Code-wise looks ok. I see it is storing the ident along with the host. Do you intend it to work this way?

2.0/m_joinpartspam.cpp
```diff
@@ -0,0 +1,243 @@
...
+			const std::string& mask(user->MakeHost());
+			if (jpss->zapme(mask))
+			{
+				std::vector<std::string> parameters;
+				parameters.push_back(chan->name);
+				parameters.push_back("+b");
+				parameters.push_back(user->MakeWildHost());
+				ServerInstance->SendGlobalMode(parameters, ServerInstance->FakeClient);
+
+				user->WriteNumeric(474, "%s %s :Channel JoinPart Spam triggered (limit is %u cycles in %u secs)", user->nick.c_str(), chan->name.c_str(), jpss->cycles, jpss->secs);
```
attilamolnar: Can you reword this to say "join/part spam" to look more natutal instead of "JoinPart Spam"?

genius3000:
@attilamolnar yes, I intended the use of ident@host; Figured host may be too broad, and using nickname as well is too specific/bypassed.
I'll change the wording in the three Numerics, that does look better.
Good to squash the commits down to just one then as well?

attilamolnar:
Ok.

> Good to squash the commits down to just one then as well?

Yeah please do

genius3000:
@atillamolnar Alright, squashed down to the one commit. Thank you very much for all the assistance.

attilamolnar:
Thanks for your contribution!

PR #91, opened 2016-09-18
Fix some issues with my 2.0 modules' directives.

SaberUK:
- Add missing ModAuthorMail directives.
- Fix core version constraints in ModDepends.


attilamolnar:
Thanks!

PR #92, opened 2016-10-31
m_ascii

Adam-:


attilamolnar:
LGTM

2.0/m_ascii.cpp
```diff
@@ -0,0 +1,55 @@
...
+/* $ModDesc: Implements CASEMAPPING=ascii */
+/* $ModDepends: core 2.0 */
+/* $ModConflicts: m_nationalchars.so */
+
+class ModuleASCII : public Module
+{
+ public:
+	void init()
+	{
+		national_case_insensitive_map = ascii_case_insensitive_map;
```
SaberUK: This might have been replaced by another module. You should probably store the original value in a private variable and reset it to that in the dtor.

Adam-: What if the original value points to the address space of another module?

attilamolnar: The problem is if we end up writing a dangling pointer there because of an unloaded module. The current behavior is the safest. Incompatibility with other modules changing the mapping is fine, it doesn't make sense to use two or more casemapping providers at once.

By the way all modules are in the same address space.

PR #94, opened 2016-11-22
Update 2.0/m_joinpartspam: improve tags, add redirect channel capability

genius3000:
- Improved the module tags and added a short documentation and helpop config

- Added the capability to have a configured redirect channel used in the ban, if m_banredirect is loaded.
  - Checks for m_banredirect and channel validity on Rehash, announces any issues.
  - Checks again upon placing a ban and doesn't try to place a redirect ban if there's an issue.
  - Is selectable using the '*' flag in setting the mode.

### DIFF
--- a/2.0/m_joinpartspam.cpp
+++ b/2.0/m_joinpartspam.cpp
@@ -16,20 +16,29 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "inspircd.h"
-
-/* $ModDesc: Adds channel mode +V to ban a user after x per y joins and parts/quits (join/part spam) */
 /* $ModAuthor: genius3000 */
 /* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Adds channel mode +V to ban a user after x per y joins and parts/quits (join/part spam) */
 /* $ModDepends: core 2.0 */
+/* $ModConfig: <joinpartspam redirchan="#BeGoneAnnoyingCycler"> */
 
+/* Helpop Lines for the CHMODES section
+ * Find: '<helpop key="chmodes" value="Channel Modes'
+ * Place just above the 'g <mask>' line:
+ V [*]<cycles>:<sec> Bans a user after the set number of Join and
+                    Part/Quit cycles in the set duration. With '*',
+                    a network configured redirect channel is used.
+ */
 /* Special thanks to Attila for the patience and guidance in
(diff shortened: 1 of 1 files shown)


2.0/m_joinpartspam.cpp
```diff
@@ -188,6 +199,21 @@ class ModuleJoinPartSpam : public Module
...
 
+	void OnRehash(User*)
+	{
+		/* Check for a channel redirect value and send an Announcement SNOTICE
+                 * if it's invalid or the ban redirect module is not loaded */
+		ConfigTag* tag = ServerInstance->Config->ConfValue("joinpartspam");
+		const std::string redirchan = tag->getString("redirchan");
+		if (!redirchan.empty())
+		{
+			if (!ServerInstance->IsChannel(redirchan.c_str(), ServerInstance->Config->Limits.ChanMax))
```
SaberUK: Rather than sending a snotice here and in the line below you would probably be better off throwing ModuleException to prevent the module from loading.

2.0/m_joinpartspam.cpp
```diff
@@ -188,6 +199,21 @@ class ModuleJoinPartSpam : public Module
 		ServerInstance->Modules->SetPriority(this, I_OnUserJoin, PRIORITY_LAST);
 	}
 
+	void OnRehash(User*)
+	{
+		/* Check for a channel redirect value and send an Announcement SNOTICE
+                 * if it's invalid or the ban redirect module is not loaded */
+		ConfigTag* tag = ServerInstance->Config->ConfValue("joinpartspam");
+		const std::string redirchan = tag->getString("redirchan");
```
genius3000: Good call, this simplifies things later on too.

2.0/m_joinpartspam.cpp
genius3000: True, if they configure a channel, they imply they want to use it. What's recommended practice for a live rehash and invalid parameter? 1) Store the old string if there is one, go back to using it; 2) Clear the string and therefore it won't be used later (at the .empty() check pre-ban setting).

SaberUK: Store the settings in local variables and once you have done all of the validation swap them with the instance variables.

genius3000:
Going to close this PR as I'm in the process of rewriting this module to block rather than ban. This is a cleaner method and allows redirect without relying on another module.

PR #95, opened 2016-12-13
m_antirandom updates

linuxdaemon:
This update includes:
-Fixing an issue with dereferencing and subscripting a pointer
-Removed redundant code
-Cleaned up the consonant/vowel switch statements
-Modified the triples code to use a hashmap

From my benchmarking, this has improved the score calculation time tremendously, only requiring about 1/4th of the time on average

### DIFF
--- a/2.0/m_antirandom.cpp
+++ b/2.0/m_antirandom.cpp
@@ -646,7 +646,7 @@ class ModuleAntiRandom : public Module
                 score++;
                 if (this->DebugMode)
                     ServerInstance->SNO->WriteGlobalSno('a', "m_antirandom: %s:MATCH triple (%s:%c/%c/%c)",
-                            original_str.c_str(), trip->second.c_str(), original_str[pos], original_str[pos + 1], original_str[pos + 2]);
+                            original_str.c_str(), trip->second.c_str(), original_str[i], original_str[i + 1], original_str[i + 2]);
             }
         }
 		return score;
--- a/2.0/m_antirandom.cpp
+++ b/2.0/m_antirandom.cpp
@@ -603,19 +603,19 @@ class ModuleAntiRandom : public Module
 
 		if (digits >= 5)
 		{
-			score += 5 + (digits - 5); // TODO is the subtraction really necessary?
+			score += digits;
 			if (this->DebugMode)
 				ServerInstance->SNO->WriteGlobalSno('a', "m_antirandom: %s:MATCH digits", original_str);
 		}
 		if (vowels >= 4)
 		{
-			score += 4 + (vowels - 4);
+			score += vowels;
 			if (this->DebugMode)
 				ServerInstance->SNO->WriteGlobalSno('a', "m_antirandom: %s:MATCH vowels", original_str);
 		}
 		if (consonants >= 4)
 		{
-			score += 4 + (consonants - 4);
+			score += consonants;
 			if (this->DebugMode)
 				ServerInstance->SNO->WriteGlobalSno('a',  "m_antirandom: %s:MATCH consonants", original_str);
 		}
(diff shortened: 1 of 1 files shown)


2.0/m_antirandom.cpp
```diff
@@ -550,9 +568,9 @@ class ModuleAntiRandom : public Module
 		unsigned int digits = 0;
 
 		/* Fast digit/consonant/vowel checks... */
-		for (s = original_str; *s; s++)
+		for (const char &c : original_str)
```
SaberUK: InspIRCd is not built with -std=c++11 and on some compilers there may be an ABI difference between -std=c++03 and -std=c++11. Can you change this to a pre-C++11 for loop?

linuxdemon1:
Ok, the for loop has been reverted and a small crash fix applied as well with strings shorter than 3 characters.

2.0/m_antirandom.cpp
```diff
@@ -510,6 +517,19 @@ static const char *triples_txt[] = {
 	NULL, NULL
 };
 
+static std::map<std::string, std::string> init_map(const char **a)
+{
+	std::map<std::string, std::string> m;
+	const char **ci;
+	for (ci = a; *ci; ci += 2)
```
attilamolnar: Please move `const char** ci` in the initialization section of the `for`.

2.0/m_antirandom.cpp
```diff
@@ -550,9 +567,10 @@ class ModuleAntiRandom : public Module
 		unsigned int digits = 0;
 
 		/* Fast digit/consonant/vowel checks... */
-		for (s = original_str; *s; s++)
+		for (i = 0; i < original_str.length(); i++)
```
attilamolnar: If `i` is only used in the loop restrict the scope of it to the loop by declaring it in the init section of the `for`

2.0/m_antirandom.cpp
```diff
@@ -564,52 +582,27 @@ class ModuleAntiRandom : public Module
...
-						highest_vowels = vowels;
-					vowels = 0;
-					break;
+				vowels++;
+			}
+			else
+			{
+				if (vowels > highest_vowels)
+					highest_vowels = vowels;
+				vowels = 0;
```
attilamolnar: I suspect the previous code with `switch` is more optimal than this due to the compiler being able to generate e.g. a jump table but I wouldn't bet on it being able to do the same with `strchr()`.

2.0/m_antirandom.cpp
```diff
@@ -623,56 +616,43 @@ class ModuleAntiRandom : public Module
...
-			 * ..run it through each triple.
-			 */
-			for (ci = triples_txt; *ci; ci++)
+			for (i = 0; i < (original_str.length() - 2); i++)
 			{
-				// At this point, ci[0] and ci[1] point to the first two chars in the triples array.
-				if (*ci[0] == s[0] && *ci[1] == s[1] && s[2])
+				std::map<std::string, std::string>::const_iterator trip;
+				// Check whether the current and next characters are the first half of a triple, if so, check for the 3rd character in the second half
+				if ((trip = triples_map.find(original_str.substr(i, 2))) != triples_map.end() &&
```
attilamolnar: Assign the value to the iterator on the previous line where the variable is declared

2.0/m_antirandom.cpp
linuxdemon1: It appears you are right, I benchmarked the `switch`-`case` vs `strchr()` over ~25 million runs, and there was an improvement when using the `switch`-`case`, but it appears to only be ~0.114 times as fast, and as such, I'm not sure how significant of a difference that would be to warrant a 20 line `switch`-`case`.

attilamolnar: Can you publish the code used in your benchmarks?

linuxdemon1: here is the code: https://gist.github.com/linuxdemon1/775890a8924b3f0972a47a1ce705ed8f

Adam-:
I think this is probably sufficient for what we need now, can revisit later if performance is that much of an issue, but the PR fixes other larger outstanding issues.

PR #99, opened 2017-02-26
Add MD5 and SHA-3 support to m_hash_gnutls.

SaberUK:
As requested by @Techman-.


attilamolnar:
There's a reason I haven't added md5 to this module when I wrote it, and that reason is the deprecated cloaking methods in m_cloaking not working with it if memory serves right. Can you verify it?

attilamolnar:
Interface in question used by m_cloaking: https://github.com/inspircd/inspircd/blob/insp20/src/modules/hash.h#L55

SaberUK:
You're probably correct, I have removed MD5 from the PR.

Techman-:
Is there a reason why that interface can't be implemented into the module?

attilamolnar:
@Techman- Yes, there are not provided by gnutls.

Techman-:
Are the already-deprecated cloaking methods going to be removed in 3.0 anyways? Seems like a crutch to keep supporting it at this point.

attilamolnar:
I agree, they have been removed a long time ago.

attilamolnar:
@SaberUK Has this been tested and confirmed to be working correctly?

SaberUK:
@attilamolnar Yes.

PR #100, opened 2017-02-28
Create m_stun.cpp

ddan39:
Provides the /STUN command, which stops a user from receiving messages/notices to channels

### DIFF
--- a/2.0/m_stun.cpp
+++ b/2.0/m_stun.cpp
@@ -1,0 +1,332 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Modified (:D) 2017 Dan39 <ddan39@gmail.com>
+ *   Copyright (C) 2009 Daniel De Graaf <danieldg@inspircd.org>
+ *   Copyright (C) 2008 Robin Burchell <robin+git@viroteck.net>
+ *   Copyright (C) 2008 Craig Edwards <craigedwards@brainbox.cc>
+ *   Copyright (C) 2008 Thomas Stagner <aquanight@inspircd.org>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+#include "xline.h"
+
+/* $ModDesc: Provides the /STUN command, which stops a user from receiving messages/notices to channels */
+
+class Stun : public XLine
+{
+ public:
+	std::string matchtext;
+
+	Stun(time_t s_time, long d, const std::string& src, const std::string& re, const std::string& stunmask)
+		: XLine(s_time, d, src, re, "STUN")
+		, matchtext(stunmask)
+	{
+	}
+
+	~Stun()
+	{
+	}
+
+	bool Matches(User *u)
+	{
+		// E: overrides stun
+		if (u->exempt)
+			return false;
+
+		if (InspIRCd::Match(u->GetFullHost(), matchtext) || InspIRCd::Match(u->GetFullRealHost(), matchtext) || InspIRCd::Match(u->nick+"!"+u->ident+"@"+u->GetIPString(), matchtext))
+			return true;
+
+		return false;
+	}
+
+	bool Matches(const std::string &s)
+	{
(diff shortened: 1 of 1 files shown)


SaberUK:
This probably should be a user mode rather than an XLine.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+ public:
+	std::string matchtext;
+
+	Stun(time_t s_time, long d, const std::string& src, const std::string& re, const std::string& stunmask)
+		: XLine(s_time, d, src, re, "STUN")
+		, matchtext(stunmask)
+	{
+	}
+
+	~Stun()
```
SaberUK: You can delete empty blocks like this.

ddan39:
I purposely made an XLine instead of a umode because I do not want the user being notified about it or able to remove it themselves. There is already the +d deaf mode which does that... The use for this module would be for me to use on people who are on my network just to spy on us, and I do not want to provoke them with a zline.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+#include "xline.h"
+
+/* $ModDesc: Provides the /STUN command, which stops a user from receiving messages/notices to channels */
```
SaberUK: You are missing a core ModDepends, and ModConfig (see what other modules do).

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+	{
+	}
+
+	bool Matches(User *u)
+	{
+		// E: overrides stun
+		if (u->exempt)
+			return false;
+
+		if (InspIRCd::Match(u->GetFullHost(), matchtext) || InspIRCd::Match(u->GetFullRealHost(), matchtext) || InspIRCd::Match(u->nick+"!"+u->ident+"@"+u->GetIPString(), matchtext))
```
SaberUK: You can just make this into a single return instead of using an if block.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+			return true;
+
+		return false;
+	}
+
+	bool Matches(const std::string &s)
+	{
+		if (matchtext == s)
```
SaberUK: `return matchtext == s;`

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+		return new Stun(set_time, duration, source, reason, xline_specific_mask);
+	}
+
+	bool AutoApplyToUserList(XLine *x)
+	{
+		return false;
+	}
+};
+
+//typedef std::vector<Stun> stunlist;
```
SaberUK: You should probably delete unused code like this.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+				user->WriteServ("NOTICE %s :*** Stun for %s already exists", user->nick.c_str(), target.c_str());
+				return CMD_FAILURE;
+			}
+		}
+		return CMD_SUCCESS;
+	}
+
+	RouteDescriptor GetRouting(User* user, const std::vector<std::string>& parameters)
+	{
+		if (IS_LOCAL(user))
```
SaberUK: `return IS_LOCAL(user) ? ROUTE_LOCALONLY : ROUTE_BROADCAST`

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+	virtual ~ModuleStun()
+	{
+		ServerInstance->XLines->DelAll("STUN");
+		ServerInstance->XLines->UnregisterFactory(&f);
+	}
+
+	/*
+	 * I dont think i need this for OnUserPreNotice and OnUserPreMessage...?
+	 *
```
SaberUK: If you don't need it then delete it.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+	void Prioritize()
+	{
+		Module* alias = ServerInstance->Modules->Find("m_alias.so");
+		ServerInstance->Modules->SetPriority(this, I_OnPreCommand, PRIORITY_BEFORE, &alias);
+	}
+	*/
+
+	virtual ModResult OnStats(char symbol, User* user, string_list& out)
+	{
+		if (symbol != 'K')
```
SaberUK: You should change this to something more unique.

2.0/m_stun.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+		deaf_bypasschars = tag->getString("bypasschars");
+		deaf_bypasschars_uline = tag->getString("bypasscharsuline");
+
+		affectopers = tag->getBool("affectopers", false);
+	}
+
+	/*
+	 * The rest of the code is mostly taken from m_deaf.so
+	 * Copyright (C) 2006-2007 Dennis Friis <peavey@inspircd.org>
```
SaberUK: If the copyright  is already mentioned in the header you don't need to repeat it.

Issue #20, opened 2013-04-26
[SUGGESTION]: update m_rmtkl


bandicot:
update the old extra module m_rmtkl.so ? , it´s so ufefull on network are atacked for proxys ect. Regards and thanks

SaberUK:
I think the behaviour of this has been included in findxline as part of #102. If this is not the case then please reopen this issue.

PR #102, opened 2017-03-26
Modify m_findxline.cpp to add Xline removal ability, reason matching, and helpop config

genius3000:
This PR adds Xline removal ~~ability~~ command and reason matching to m_findxline.cpp as well as helpop config.
~~The command now takes a '-' parameter before the Xline type (or *), which will remove all matching Xlines. The command also accepts an optional 'reason' string, '*' is used if not specified.~~
There is now a second command, '/RMXLINE' which will remove matching Xlines, same syntax as '/FINDXLINE'
Both commands do ban mask and reason matching (reason is optional and will be '*' if not specified).

Testing has been performed on a two server + services test network with removal of all Xlines of a certain type, matching Xlines of a certain type, and matching Xlines of all types (both find and rm).

This PR technically brings the removal functionality of m_rmtkl.cpp (from 1.2/) into m_findxline.cpp (Issue #20)

### DIFF
--- a/2.0/m_findxline.cpp
+++ b/2.0/m_findxline.cpp
@@ -1,6 +1,7 @@
 /*
  * InspIRCd -- Internet Relay Chat Daemon
  *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
  *   Copyright (C) 2013 Attila Molnar <attilamolnar@hush.com>
  *
  * This file is part of InspIRCd.  InspIRCd is free software: you can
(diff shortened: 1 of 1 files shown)


SaberUK:
Would a separate command not be more useful for this?

genius3000:
Yes, yes it would be. I'll make this change, a separate remove command which won't re-list the Xlines. Thanks.

Cronus89:
Exactly what I wanted, can this be merged so I can use it in modulemanager?

2.0/m_findxline.cpp
```diff
@@ -1,6 +1,9 @@
 /*
  * InspIRCd -- Internet Relay Chat Daemon
  *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
+ *                 - XLine removal command
+ *                 - Reason matching
```
SaberUK: There's no need to put what you changed here. Thats what git history is for.

2.0/m_findxline.cpp
```diff
@@ -50,26 +72,119 @@ class CommandFindXLine : public Command
...
+			for (std::vector<std::string>::iterator i = xlinetypes.begin(); i != xlinetypes.end(); ++i)
+			{
+				XLineLookup* xlines = ServerInstance->XLines->GetAll(*i);
+				if (xlines)
+					List(user, parameters[1], reason, *i, xlines, matched, total);
+			}
+		}
+		else
+		{
+			std::string linetype = parameters[0];
```
SaberUK: `const std::string&`

PR #103, opened 2017-03-31
Adding m_conn_vhost: allows configuring a vhost in a connect block.

genius3000:
This allows configuring a vhost per connect block, this may be wanted over normal cloaking or configured/account based vhosting. The module m_hostchange could handle this want in some scenarios but it has a major downfall currently, you can only match to a hostmask once. So you cannot match to '*' multiple times on different ports (possibly simulating a connect block criteria).
The want/request for this feature can be seen in this [issue](https://github.com/inspircd/inspircd/issues/1313) in the 'inspircd' repo.

### DIFF
--- a/2.0/m_conn_vhost.cpp
+++ b/2.0/m_conn_vhost.cpp
@@ -1,0 +1,90 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Adds a "vhost" option to connect blocks */
+/* $ModDepends: core 2.0 */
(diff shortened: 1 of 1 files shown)


2.0/m_conn_vhost.cpp
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	void Prioritize()
+	{
+		/* Let's go after conn_umodes in case +x also gets set */
+		Module* connumodes = ServerInstance->Modules->Find("m_conn_umodes.so");
+		ServerInstance->Modules->SetPriority(this, I_OnUserConnect, PRIORITY_AFTER, connumodes);
+	}
+
+	virtual Version GetVersion()
```
SaberUK: This + OnUserConnect don't need to be virtual.

2.0/m_conn_vhost.cpp
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	virtual void OnUserConnect(LocalUser* user)
+	{
+		ConfigTag* tag = user->MyClass->config;
+		std::string vhost = tag->getString("vhost");
+
+		if (!vhost.empty())
+		{
+				user->ChangeDisplayedHost(vhost.c_str());
```
SaberUK: Your indentation is kinda iffy here.

2.0/m_conn_vhost.cpp
```diff
@@ -0,0 +1,60 @@
...
+	{
+		return Version("Sets a connect block configured vhost on users when they connect");
+	}
+
+	virtual void OnUserConnect(LocalUser* user)
+	{
+		ConfigTag* tag = user->MyClass->config;
+		std::string vhost = tag->getString("vhost");
+
+		if (!vhost.empty())
```
SaberUK: You might want to do some validation here to check its actually a valid host.

2.0/m_conn_vhost.cpp
genius3000: Not kinda, it was one too many tabs...totally missed that.

2.0/m_conn_vhost.cpp
genius3000: Not a bad idea, I was going for the "less is more" style with the simplicity here but ensuring we don't crash the server on over-length or an invalid character is probably a good plan.

SaberUK:
LGTM.

PR #104, opened 2017-04-28
Backport m_blockhighlight from the development branch.

SaberUK:
The following new config settings have also been added:
* `<blockhighlight:stripcolor>` - whether to strip colour codes before matching
* `<blockhighlight:reason>` - the message to kill spamming clients with

### DIFF
--- a/2.0/m_blockhighlight.cpp
+++ b/2.0/m_blockhighlight.cpp
@@ -1,0 +1,119 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 Peter Powell <petpow@saberuk.com>
+ *   Copyright (C) 2015 Attila Molnar <attilamolnar@hush.com>
+ *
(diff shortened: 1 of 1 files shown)


Derecho:
Hi, this looks really useful, I'm considering running this when it gets into master.

What about adding bans/glines as well though, apart from just the kill? This could be an optional parameter when configuring the module.

2.0/m_blockhighlight.cpp
```diff
@@ -0,0 +1,118 @@
...
+		// Must be at least minlen long
+		if (text.length() < minlen)
+			return MOD_RES_PASSTHRU;
+
+		Channel* const chan = static_cast<Channel*>(dest);
+		if (chan->GetUsers()->size() < minusers)
+			return MOD_RES_PASSTHRU;
+
+		if (stripcolor)
+			InspIRCd::StripColor(text);
```
Adam-: This should maybe strip a copy of the message?

PR #107, opened 2017-07-05
Updates to the blockhighlight module.

SaberUK:
- Add a new <blockhighlight:ignoreextmsg> setting to prevent channel membership enumeration.
- Update the default reason to be more user friendly.
- Make the channel mode not completely useless (whoops lol, not sure how my testing didn't catch that).

### DIFF
--- a/2.0/m_blockhighlight.cpp
+++ b/2.0/m_blockhighlight.cpp
@@ -19,7 +19,7 @@
 
 /* $ModAuthor: Peter "SaberUK" Powell */
 /* $ModAuthorMail: petpow@saberuk.com */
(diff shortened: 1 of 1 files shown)


2.0/m_blockhighlight.cpp
```diff
@@ -74,6 +77,14 @@ class ModuleBlockHighlight : public Module
 		if (chan->GetUsers()->size() < minusers)
 			return MOD_RES_PASSTHRU;
 
+		// We only work if the channel mode is enabled.
+		if (!chan->IsModeSet(mode.GetModeChar()))
+			return MOD_RES_PASSTHRU;
```
SISheogorath: Is this the most efficient order? I think we should check if the mode is active before we check the channel size and the message length.

SaberUK: It doesn't really matter that much.

PR #108, opened 2017-07-13
Add feedback message when reloading SSL certificates

Brainscrewer:
Currently you do not get any feedback message on the IRC server itself
when using kill -SIGUSR1 <PID> when reloading the SSL certificates. This
patch adds the feedbackmessage so you will see this message if you have
snomasks enabled, or you have a channel which logs snomask messages.

### DIFF
--- a/2.0/m_rehashsslsignal.cpp
+++ b/2.0/m_rehashsslsignal.cpp
@@ -50,7 +50,9 @@ class ModuleRehashSSLSignal : public Module
 		if (!signaled)
 			return;
 
-		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, "Got SIGUSR1, reloading SSL credentials");
+		const std::string feedbackmsg = "Got SIGUSR1, reloading SSL credentials";
+		ServerInstance->SNO->WriteGlobalSno('a', feedbackmsg);
+		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, feedbackmsg);
 		const std::string str = "ssl";
 		FOREACH_MOD(I_OnModuleRehash, OnModuleRehash(NULL, str));
 		signaled = 0;


2.0/m_rehashsslsignal.cpp
```diff
@@ -50,6 +50,9 @@ class ModuleRehashSSLSignal : public Module
 		if (!signaled)
 			return;
 
+		std::string m = "SSL credentials have been reloaded via terminal on " + ServerInstance->Config->ServerName ;
```
genius3000: Don't need the space before the ';'

2.0/m_rehashsslsignal.cpp
```diff
@@ -50,6 +50,9 @@ class ModuleRehashSSLSignal : public Module
...
+		ServerInstance->SNO->WriteGlobalSno('a', m);
+		
 		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, "Got SIGUSR1, reloading SSL credentials");
```
genius3000: Could probably use this existing message for both Log and SNOTICE. No need to say what server in the message as the SNOTICE will come from that server.

2.0/m_rehashsslsignal.cpp
```diff
@@ -50,7 +50,10 @@ class ModuleRehashSSLSignal : public Module
 		if (!signaled)
 			return;
 
-		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, "Got SIGUSR1, reloading SSL credentials");
+		std::string m = "Got SIGUSR1, reloading SSL credentials";
```
genius3000: Should be `const std::string`

2.0/m_rehashsslsignal.cpp
```diff
@@ -50,7 +50,10 @@ class ModuleRehashSSLSignal : public Module
...
+		ServerInstance->SNO->WriteGlobalSno('a', m);
+
```
genius3000: (Nitpicking here but) this line break might look better between the Log line and `const std::string str...`
ie. keep the three lines dealing with the message together.

2.0/m_rehashsslsignal.cpp
SaberUK: Also, give it a better name than `m`.

Brainscrewer:
@genius3000 made the changes.

genius3000:
@Brainscrewer looks good, now you should squash the 3 commits into 1. Your first commit message looks good to keep, then this is merge-able imo.

attilamolnar:
> Currently you do not get any feedback message on the IRC server itself
when using kill -SIGUSR1 when reloading the SSL certificates.

This is a deliberate omission in my eyes. The event is logged locally, this can be useful if a cronjob reloads the cert every day and wants to verify the server's reaction. In this use case, with a dozen servers a global announcement from each would be annoying, especially as it does not say anything about the reload being successful or not.

Can you please elaborate why is a global message needed?

Brainscrewer:
@attilamolnar Ah, i didn't really think of it that way. The reason I wanted to implement this is that, since we disabled file-based logging on the server, we still didn't get a feedback message regarding the reloading of the certificates. Since we are using m_chanlog to send SNOMASKS to a channel we now would be able to get some kind of feedback message.

genius3000:
I see @attilamolnar 's point, but imagine a lot of use cases would be 1-4 times a month reloading the certificates.
With chanlog enabled on 'a', you wouldn't need to send the SNOTICE globally.

Maybe a good in between is a config boolean for a local SNOTICE or none.

Brainscrewer:
@genius3000 in regards to squashing: isn't that something that is available now per github repo? https://github.com/blog/2141-squash-your-commits

genius3000:
@Brainscrewer I did not know about that feature (I've never merged any PRs on GH). I know in the past inspircd/ has preferred additions like this to have small corrections squashed.
One of the repo maintainers can say which method they'd prefer.

2.0/m_rehashsslsignal.cpp
```diff
@@ -50,7 +50,10 @@ class ModuleRehashSSLSignal : public Module
 		if (!signaled)
 			return;
 
-		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, "Got SIGUSR1, reloading SSL credentials");
+		const std::string feedbackmsg = "Got SIGUSR1, reloading SSL credentials";
+		ServerInstance->SNO->WriteGlobalSno('a', feedbackmsg);
+		ServerInstance->Logs->Log("m_rehashsslsignal", DEFAULT, feedbackmsg);
+
```
SaberUK: There is unnecessary whitespace on this line.

Brainscrewer:
@attilamolnar @SaberUK Will something be done with this? It's been almost 2 months.

PR #114, opened 2017-10-01
Add m_conn_matchident: match by ident in connect class

genius3000:
I know that matching purely by ident in a connect class can be risky and is likely not a common use-case (looking at you @Cronus89 :smiling_imp:). That said, here's a module that can do just that.  
(Edit): The first version of this commit had incorrect logic in place, I've rewrote that to be a lot cleaner and safer now.

### DIFF
--- a/2.0/m_conn_matchident.cpp
+++ b/2.0/m_conn_matchident.cpp
@@ -1,0 +1,59 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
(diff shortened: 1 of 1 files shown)


2.0/m_conn_matchident.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Adds a "matchident" option to connect blocks */
```
SaberUK: This description isn't really very useful. Maybe use the same as in GetVersion?

genius3000: Yeah, forgot to change that after; fixed now.

SaberUK:
LGTM. I'll leave it open for a few days to let people other comment. Poke me if I haven't merged by Friday though.

2.0/m_conn_matchident.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "inspircd.h"
+
+class ModuleConnMatchIdent : public Module
+{
+ public:
+	void init()
+	{
+		Implementation eventlist[] = { I_OnSetConnectClass };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
```
SaberUK: If you're only hooking to one event you can just use the non-array form of Attach.

genius3000: Good point, fixed.

PR #115, opened 2017-10-14
Add m_checkbans

genius3000:
Provides three commands: /checkbans, /testban, and /whyban
`checkbans`: Check a channel's ban/exception list against its current users
`testban`: Test a ban mask against the current users in a channel
`whyban`: See what bans/exceptions match yourself or another user to a channel

I'm thinking the numerics currently used should be changed, as some clients probably won't like seeing the Banlist and End of Banlist numerics used here. Any suggestions?

### DIFF
--- a/2.0/m_checkbans.cpp
+++ b/2.0/m_checkbans.cpp
@@ -1,0 +1,274 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Adds commands /checkbans, /testban, and /whyban */
+/* $ModDepends: core 2.0 */
+
+/* Helpop Lines for the CUSER section
+ * Find: '<helpop key="cuser" value="User Commands
+ * Place 'TESTBAN  WHYBAN    CHECKBANS' after 'SSLINFO'
+ * Re-space as needed to match the current columns
+ * Find: '<helpop key="sslinfo" ...'
+ * Place just above that line:
+<helpop key="testban" value="/TESTBAN <channel> <mask>
+
+Test a channel ban mask against the users currently in the channel.">
+
+<helpop key="whyban" value="/WHYBAN <channel> [user]
+
+Get a list of bans and exceptions that match you (or the given user)
+on the specified channel.">
+
+<helpop key="checkbans" value="/CHECKBANS <channel>
+
+Get a list of bans and exceptions that match current users on the channel.">
+
+ */
+
+
+#include "inspircd.h"
+#include "u_listmode.h"
+
+/* Common function */
+bool CanCheck(Channel* c, User* u)
+{
+	if (u->HasPrivPermission("channels/auspex"))
+		return true;
+
+	ModeHandler* mh = ServerInstance->Modes->FindMode('b', MODETYPE_CHANNEL);
+	return (mh->GetLevelRequired() <= c->GetPrefixValue(u));
+}
+
+/* Handle /CHECKBANS */
+class CommandCheckBans : public Command
+{
+ public:
+	CommandCheckBans(Module* Creator) : Command(Creator, "CHECKBANS", 1, 1)
+	{
+		this->syntax = "<channel>";
+	}
+
+	CmdResult Handle(const std::vector<std::string>& parameters, User* user)
+	{
+		Channel* c = ServerInstance->FindChan(parameters[0]);
+		if (!c)
+		{
+			user->WriteNumeric(ERR_NOSUCHNICK, "%s %s :No such channel", user->nick.c_str(), parameters[0].c_str());
+			return CMD_FAILURE;
+		}
+
+		/* Only allow checking for matching users if you have access to the ban list. */
+		if (!CanCheck(c, user))
+		{
+			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, "%s %s :You do not have access to the ban list.",
+				user->nick.c_str(), c->name.c_str());
+			return CMD_FAILURE;
+		}
+
+		/* Check for channel mode 'e', ban exceptions */
+		ListModeBase* ex = (ListModeBase*)ServerInstance->Modes->FindMode('e', MODETYPE_CHANNEL);
+		modelist* exceptions = NULL;
+		if (ex)
+			exceptions = ex->extItem.get(c);
+
+		/* Loop through all users of the channel, checking for matches to bans and exceptions */
+		const UserMembList* users = c->GetUsers();
+		for (UserMembCIter u = users->begin(); u != users->end(); u++)
+		{
+			for (BanList::iterator ban = c->bans.begin(); ban != c->bans.end(); ban++)
+			{
+				if (c->CheckBan(u->first, ban->data))
+				{
+					user->WriteServ("367 %s %s Ban: %s matches %s (set by: %s on %s)", user->nick.c_str(),
+						c->name.c_str(), ban->data.c_str(), u->first->nick.c_str(),
+						ban->set_by.c_str(), ServerInstance->TimeString(ban->set_time).c_str());
+				}
+			}
+
+			if (!exceptions)
+				continue;
+
+			for (modelist::iterator exc = exceptions->begin(); exc != exceptions->end(); exc++)
+			{
+				if (c->CheckBan(u->first, exc->mask))
+				{
+					user->WriteServ("367 %s %s Exception: %s matches %s (set by: %s on %s)", user->nick.c_str(),
+						c->name.c_str(), exc->mask.c_str(), u->first->nick.c_str(), exc->nick.c_str(),
+						ServerInstance->TimeString(strtoul(exc->time.c_str(), NULL, 0)).c_str());
+				}
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+		/* Only allow checking for matching users if you have access to the ban list. */
+		if (!CanCheck(c, user))
+		{
+			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, "%s %s :You do not have access to the ban list.",
+				user->nick.c_str(), c->name.c_str());
+			return CMD_FAILURE;
+		}
+
+		/* Check for channel mode 'e', ban exceptions */
+		ListModeBase* ex = (ListModeBase*)ServerInstance->Modes->FindMode('e', MODETYPE_CHANNEL);
```
SaberUK: You should check the mode name here before casting it as 'e' might be bound to another mode.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+		modelist* exceptions = NULL;
```
SaberUK: This could just be something like

```cpp
modelist* exceptions = ex ? ex->extItem->get(c) : NULL;
```

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+		/* Loop through all users of the channel, checking for matches to bans and exceptions */
+		const UserMembList* users = c->GetUsers();
+		for (UserMembCIter u = users->begin(); u != users->end(); u++)
+		{
+			for (BanList::iterator ban = c->bans.begin(); ban != c->bans.end(); ban++)
+			{
+				if (c->CheckBan(u->first, ban->data))
+				{
+					user->WriteServ("367 %s %s Ban: %s matches %s (set by: %s on %s)", user->nick.c_str(),
```
SaberUK: 1. This numeric collides with other numerics already used by InspIRCd. Pick something else that [is not already in use](https://defs.ircdocs.horse/defs/numerics.html).

2. This is not a well formed IRC message. You should prefix the final text parameter with a `:` to make it into a <trailing> parameter.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+						c->name.c_str(), ban->data.c_str(), u->first->nick.c_str(),
+						ban->set_by.c_str(), ServerInstance->TimeString(ban->set_time).c_str());
+				}
+			}
+
+			if (!exceptions)
```
SaberUK: If `exceptions` is not used until here then you should bring it down here instead of leaving it up top.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+				continue;
+
+			for (modelist::iterator exc = exceptions->begin(); exc != exceptions->end(); exc++)
+			{
+				if (c->CheckBan(u->first, exc->mask))
+				{
+					user->WriteServ("367 %s %s Exception: %s matches %s (set by: %s on %s)", user->nick.c_str(),
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+						c->name.c_str(), exc->mask.c_str(), u->first->nick.c_str(), exc->nick.c_str(),
+						ServerInstance->TimeString(strtoul(exc->time.c_str(), NULL, 0)).c_str());
+				}
+			}
+		}
+
+		user->WriteServ("368 %s %s :End of check bans list", user->nick.c_str(), c->name.c_str());
```
SaberUK: The same as above re: numeric usage applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+			return CMD_FAILURE;
+		}
+
+		unsigned int matched = 0;
+		const UserMembList* users = c->GetUsers();
+		for (UserMembCIter u = users->begin(); u != users->end(); u++)
+		{
+			if (c->CheckBan(u->first, parameters[1]))
+			{
+				user->WriteServ("367 %s %s Ban: %s matches %s", user->nick.c_str(), c->name.c_str(),
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+					parameters[1].c_str(), u->first->nick.c_str());
+				matched++;
+			}
+		}
+
+		if (matched > 0)
+		{
+			float percent = ((float)matched / (float)users->size()) * 100;
+			user->WriteServ("367 %s %s Ban: %s matched %d of %lu users (%.2f%%).", user->nick.c_str(), c->name.c_str(),
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+						parameters[1].c_str(), matched, users->size(), percent);
+		}
+
+		user->WriteServ("368 %s %s :End of test ban list", user->nick.c_str(), c->name.c_str());
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+		{
+			user->WriteNumeric(ERR_NOSUCHNICK, "%s %s :No such channel", user->nick.c_str(), parameters[0].c_str());
+			return CMD_FAILURE;
+		}
+
+		/* Allow checking yourself against channel bans with no access, but only
+		 * allow checking others if you have access to the channel ban list.
+		 */
+		User* u;
+		if (parameters.size() == 1)
```
SaberUK:
```cpp
User* u == parameters.size() == 1 ? user : NULL;
```

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+			user->WriteNumeric(ERR_NOSUCHNICK, "%s %s :No such nick", user->nick.c_str(), parameters[1].c_str());
+			return CMD_FAILURE;
+		}
+
+		/* Check the ban list first */
+		for (BanList::iterator ban = c->bans.begin(); ban != c->bans.end(); ban++)
+		{
+			if (c->CheckBan(u, ban->data))
+			{
+				user->WriteServ("367 %s %s Ban: %s (set by: %s on %s)", user->nick.c_str(), c->name.c_str(),
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+					ban->data.c_str(), ban->set_by.c_str(), ServerInstance->TimeString(ban->set_time).c_str());
+			}
+		}
+
+		/* Now try to get the exceptions list and check it */
+		ListModeBase* ex = (ListModeBase*)ServerInstance->Modes->FindMode('e', MODETYPE_CHANNEL);
```
SaberUK: The same as above re: checking the mode name.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,274 @@
...
+		modelist* exceptions = NULL;
+		if (ex)
+			exceptions = ex->extItem.get(c);
+		if (exceptions)
+		{
+			for (modelist::iterator exc = exceptions->begin(); exc != exceptions->end(); exc++)
+			{
+				if (c->CheckBan(u, exc->mask))
+				{
+					user->WriteServ("367 %s %s Exception: %s (set by: %s on %s)", user->nick.c_str(),
```
SaberUK: The same as above re: numeric usage and message validity applies here too.

SaberUK:
You might want to consider checking if you can deduplicate some of the code in this as theres quite a bit which seems to be duplicated.

bigfoot547:
Could this module be used to decloak people?
If so, could there be some flood penalty applied?

genius3000:
> You might want to consider checking if you can deduplicate some of the code in this as theres quite a bit which seems to be duplicated. --@SaberUK 

I pondered that a bit too before, I'll see what I can do though. Also thanks for the reviews, I'll find unused numerics and name them; as well as clean up the messages.

> Could this module be used to decloak people?
If so, could there be some flood penalty applied? --@bigfoot547 

Yeah it could be used to decloak, limiting to halfop/op helps some but I'll add a flood penalty as well and look into any other options to limit/prevent abusing this. Thanks for the catch.

genius3000:
Alright I rebased and added a commit that should take care of all the comments/concerns.  
Managed to deduplicate the main listing code, exception mode check (also fixed the casting before knowing what it is issue), and the 'need chanops' message.  
Found unused numeric numbers to use and named them, fixed the messages, and added a penalty of 6 to both /checkbans and /testban, with a penalty of 10 to /whyban <nick>
I'll squash the commits once approved.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,263 @@
...
+			return CMD_FAILURE;
+		}
+
+		/* Allow checking yourself against channel bans with no access, but only
+		 * allow checking others if you have access to the channel ban list.
+		 */
+		User* u = parameters.size() == 1 ? user : NULL;
+		if (!u)
+		{
+			this->Penalty = 10;
```
linuxdaemon: I believe what you meant here is `user->CommandFloodPenalty += 10000;` to set a user's penalty. This will currently modify the penalty for the command if anyone attempts to check another user.

2.0/m_checkbans.cpp
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+
+		// Check for ban exceptions as mode 'e'
+		ListModeBase* ex = CheckExceptionMode();
+
+		// Check for matching bans and exceptions (if available)
+		CheckLists(user, c, u, ex);
+
+		user->WriteNumeric(RPL_ENDLIST, "%s %s :End of why ban list", user->nick.c_str(), c->name.c_str());
```
linuxdaemon: Might be better to do something like `nick channel checked_nick :End of /WHYBAN list` to mimic how commands like `WHO` respond.

2.0/m_checkbans.cpp
genius3000: If I recall correctly, that's exactly my intention. The command handler will then assign the penalty.

linuxdaemon: Yes but this currently sets the penalty for the command no matter what, after that line, any time the command is ran a penalty of 10 will be applied.

genius3000: Ah yeah I see what you're saying now. Thanks.

PR #116, opened 2017-10-22
Backport m_classban from a 3.0 vendor module to a 2.0 extras

genius3000:
Not sure if there's many requests for this module in 2.0, but it came to mind earlier today and I figured why not backport it. I based the 2.0 extban code on m_gecosban.
#### Testing
* Loaded on 1 of 3 servers, works on the 1 server as intended
* Loaded on all 3 servers, works across servers as intended

### DIFF
--- a/2.0/m_classban.cpp
+++ b/2.0/m_classban.cpp
@@ -1,0 +1,65 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 genius3000 <genius3000@g3k.solutions>
+ *   Copyright (C) 2016 Johanna Abrahamsson <johanna-a@mjao.org>
+ *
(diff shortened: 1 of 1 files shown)


2.0/m_classban.cpp
```diff
@@ -0,0 +1,66 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2016 Johanna Abrahamsson <johanna-a@mjao.org>
+ *  Backported from master (3.0) to 2.0 by genius3000, original at:
```
SaberUK: You can just put a normal copyright line here, no need to say that its a backport.

genius3000: Fixed and noted for future, thanks.

PR #117, opened 2017-12-05
Add m_timedstaticquit

bigfoot547:
This module will replace a quit message of a user who has been connected for less than a specified amount of time with a configurable static quit message.

### DIFF
--- a/2.0/m_timedstaticquit.cpp
+++ b/2.0/m_timedstaticquit.cpp
@@ -1,0 +1,77 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 bigfoot547 <bigfoot@bigfootslair.net>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: bigfoot547 */
+/* $ModAuthorMail: bigfoot@bigfootslair.net */
+/* $ModDesc: Replaces the quit message of a quitting user if they have been connected for less than a configurable time */
+/* $ModDepends: core 2.0 */
+/* $ModConfig: <timedstaticquit quitmsg="Client Quit" mintime="300"> */
+
+#include "inspircd.h"
+
+class ModuleTimedStaticQuit : public Module
+{
+ private:
+	std::string quitmsg;
+	unsigned int mintime;
+
+ public:
+	void init()
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnPreCommand, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+	}
+
+	ModResult OnPreCommand(std::string& command, std::vector<std::string>&, LocalUser* user, bool validated, const std::string&)
+	{
+		// We check if the user has done his due time on the network
+		if (validated && (command == "QUIT"))
+		{
+			time_t now = ServerInstance->Time();
+			if ((now - user->signon) < mintime)
+			{
+				ServerInstance->Users->QuitUser(user, quitmsg);
+				return MOD_RES_DENY;
+			}
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	void OnRehash(User* user)
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("timedstaticquit");
+		this->quitmsg = tag->getString("quitmsg", "Client Quit");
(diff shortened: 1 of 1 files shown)


2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnPreCommand, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string>& parameters, LocalUser *user, bool validated, const std::string &original_line)
+	{
+		// We check if the user has done his due time on the network
+		if (command == "QUIT")
```
SaberUK: This could potentially cause problems if the command has not been validated yet (see the fourth parameter).

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+public:
+	void init()
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnPreCommand, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string>& parameters, LocalUser *user, bool validated, const std::string &original_line)
```
SaberUK: &* should go against the type not the name (and if you're not using the parameter you can omit the name).

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+				return MOD_RES_DENY;
+			}
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	void OnRehash(User* user)
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("timedstaticquit");
+		this->quitmsg = tag->getString(std::string("quitmsg"), "Client Quit");
```
SaberUK: You don't need to do `std::string(...)` here. Just pass the name as a string literal.

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+/* $ModDepends: core 2.0 */
+/* $ModConfig: <timedstaticquit quitmsg="Client Quit" mintime="300"> */
+
+#include "inspircd.h"
+
+class ModuleTimedStaticQuit : public Module
+{
+private:
+	std::string quitmsg;
+	int mintime;
```
SaberUK: This should be unsigned.

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	void OnRehash(User* user)
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("timedstaticquit");
+		this->quitmsg = tag->getString(std::string("quitmsg"), "Client Quit");
+		this->mintime = tag->getInt(std::string("mintime"), 300); /* Duration is in seconds */
+		this->mintime = (this->mintime < 1) ? (1) : (this->mintime); // The minimum time needs to be at least 1 second
```
SaberUK: Perhaps doing something like this would be better:

```cpp
int duration = ServerInstance->Duration(tag->getString("mintime", "5m"));
this->mintime = duration <= 0 ? 1 : duration;
```

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	virtual Version GetVersion()
```
SaberUK: The virtual keyword is unnecessary here.

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,70 @@
...
+	{
+		return Version("Replaces the quit message of a quitting user if they have been connected for less than a configurable time", VF_OPTCOMMON); // It was late, mk? Make me a better description pls :)
```
SaberUK: Put the comment on the previous line please.

bigfoot547:
@SaberUK: I have done the stuff, can you please review again?

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+public:
+	void init()
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnPreCommand, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+	}
+
+	ModResult OnPreCommand(std::string& command, std::vector<std::string>&, LocalUser *user, bool validated, const std::string&)
```
SaberUK: You might want to deprioritize this event so it is executed last in case any other modules change it.

Also, the pointer on user should be with the type not the name.

2.0/m_timedstaticquit.cpp
```diff
@@ -0,0 +1,71 @@
...
+/* $ModAuthorMail: bigfoot@bigfootslair.net */
+/* $ModDesc: Replaces the quit message of a quitting user if they have been connected for less than a configurable time */
+/* $ModDepends: core 2.0 */
+/* $ModConfig: <timedstaticquit quitmsg="Client Quit" mintime="300"> */
+
+#include "inspircd.h"
+
+class ModuleTimedStaticQuit : public Module
+{
+private:
```
SaberUK: There should be a single space character before private and public (below).

SaberUK:

PR #120, opened 2017-12-19
Add m_bannegate to negate bans with a ~

B00mX0r:
This module will match the opposite of any ban if the matching mask is prefixed with a ~.

### DIFF
--- a/2.0/m_bannegate.cpp
+++ b/2.0/m_bannegate.cpp
@@ -1,0 +1,56 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 Dylan Frank <b00mx0r@aureus.pw>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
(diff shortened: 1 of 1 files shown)


2.0/m_bannegate.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+ 	void Prioritize()
+ 	{
+ 		ServerInstance->Modules->SetPriority(this, I_OnCheckBan, PRIORITY_FIRST);
+ 	}
+
+	ModResult OnCheckBan(User *source, Channel *chan, const std::string& mask)
+	{
+		// If our matching mask begins with the negate charater, but does not have multiple in a row (to avoid nested loops)
+		if (mask[0] == '~' && mask[1] != '~')
```
SaberUK: You should check the length here.

2.0/m_bannegate.cpp
```diff
@@ -0,0 +1,57 @@
...
+		{
+			bool RealResult = chan->CheckBan(source, mask.substr(1));
```
SaberUK: There's no real need to store this in a temporary.

2.0/m_bannegate.cpp
```diff
@@ -0,0 +1,57 @@
...
+ 	{
+ 		ServerInstance->Modules->Attach(I_OnCheckBan, this);
+ 	}
+
+ 	void Prioritize()
+ 	{
+ 		ServerInstance->Modules->SetPriority(this, I_OnCheckBan, PRIORITY_FIRST);
+ 	}
+
+	ModResult OnCheckBan(User *source, Channel *chan, const std::string& mask)
```
SaberUK: Pointers should go against the type not the name.

2.0/m_bannegate.cpp
B00mX0r: This isn't meant to only be for extbans. You can also do +b ~nick!*@* to ban everyone except nick!*@*. While that probably won't be used much, this flexibility is why I love inspircd modules.

SaberUK: Yes but theres no guarantee that `mask[1]` will actually be set and if its not this will cause a crash.

B00mX0r: Right, thanks for catching that! Good to know.

PR #121, opened 2018-01-11
m_blockhighlight: Add operoverride setting and exemptchanops support

Techman-:
For InspIRCd 2.0


SaberUK:
Rather than an explicit bool in the config can you add an oper privilege instead?

PR #123, opened 2018-01-22
Add m_extbanregex

genius3000:
Provides extban 'x' for regex matching with a configurable engine.  
Modewatchers for core 'b' (ban), along with modules 'e' (banexception) and 'I' (inviteexception). The latter two are added/deleted dependent on the respective module status. The modewatchers ensure a regex engine is properly loaded and that the mask is a valid regex string.  
A change in regex engine will cause removal of all set regex extbans per local server.  
Syntax for the mask is the same as R-Line (nick!user@host\srealname); matching is done against three variations of host - dynamic (cloaked/v host), real host, and IP.

### DIFF
--- a/2.0/m_extbanregex.cpp
+++ b/2.0/m_extbanregex.cpp
@@ -1,0 +1,362 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Provides extban 'x' - Regex matching to n!u@h\sr */
+/* $ModDepends: core 2.0 */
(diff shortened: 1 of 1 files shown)


2.0/m_extbanregex.cpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+void ModeLine(const std::string& chan, char modechar, const std::vector<std::string>& masks)
+{
+	std::vector<std::string> modestr;
+	modestr.push_back(chan);
+	modestr.push_back("-");
+
+	// Total length of message "MODE <chan> -" so far
+	unsigned int strlength = 7 + chan.length();
```
SaberUK: Can this magic number be made into a constant?

2.0/m_extbanregex.cpp
```diff
@@ -0,0 +1,362 @@
...
+		CUList empty;
+		std::string notice = "Regex engine has changed to '" + engine + "'. All regex extbans have been removed";
+		ModeHandler* hop = ServerInstance->Modes->FindMode('h', MODETYPE_CHANNEL);
+		char pfxchar = ((hop) && (hop->name == "halfop")) ? hop->GetPrefix() : '@';
+
+		chan->WriteAllExcept(ServerInstance->FakeClient, true, pfxchar, empty, "NOTICE %s :%s", chan->name.c_str(), notice.c_str());
+	}
+}
+
+class BanWatcher : public ModeWatcher
```
SaberUK: Can all of these be made into one class which takes the mode to watch as a parameter?

genius3000: Totally never thought to do it that way, should work.

SaberUK:
Should this be only made available to opers? Ban matching is done in the main loop and therefore this could potentially kill the server if someone untrusted sets a ban that requires a lot of CPU time to match..

genius3000:
Fair point, could do a config option that defaults to opers only.

bigfoot547:
Could it send a SNO if it takes too long to match a ban? Then you could tell if someone is abusing it or not.

genius3000:
The reviews and concerns should be addressed in the latest commit. I'm not totally sure on the time check, kinda just pieced it together with some googlefu. It does work, but is there a downside to checking time twice so often? Is half a second too high or too low? I was seeing 0.000002 to 0.000023 seconds on some test regex strings with a (purposely) terrible string running a high of 1.03 seconds.

PR #124, opened 2018-01-22
Add m_extbanbanlist

genius3000:
Provides extban 'b' for including the ban list of another channel.  
You need to have half-op/op in the target channel (access to modify that ban list). ~~The ban check into the target channel will ignore any 'b:' masks to prevent multiple jumps or circular loops.~~ Recursive checking is blocked by use of a 'checking' flag.

### DIFF
--- a/2.0/m_extbanbanlist.cpp
+++ b/2.0/m_extbanbanlist.cpp
@@ -1,0 +1,129 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Provides extban 'b' - Ban list from another channel */
+/* $ModDepends: core 2.0 */
+
+/* Helpop Lines for the EXTBANS section
+ * Find: '<helpop key="extbans" value="Extended Bans'
+ * Place just before the 'j:<channel>' line:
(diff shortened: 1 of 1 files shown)


bigfoot547:
Yay, I am excited to see this feature in InspIRCd! :)

2.0/m_extbanbanlist.cpp
```diff
@@ -0,0 +1,129 @@
...
+			Channel* chan = ServerInstance->FindChan(mask.substr(2));
+			if (!chan)
+				return MOD_RES_PASSTHRU;
+
+			for (BanList::iterator ban = chan->bans.begin(); ban != chan->bans.end(); ++ban)
+			{
+				std::string& banmask = ban->data;
+
+				// Prevent multiple (and circular) banlist loops
+				if (banmask.length() > 2 && banmask[0] == 'b' && banmask[1] == ':')
```
linuxdaemon: How does this handle `U:b:#channel`? From my understanding, that would cause it to loop. Possibly setting a module-wide boolean while iterating the banlist and doing something like 
```c++
if (this->checking)
    return MOD_RES_PASSTHRU;
```
at the beginning of `OnCheckBan()` would help prevent odd cases like this.

If the above change is made, the for-loop can be replaced with a call to `chan->IsBanned(user)` which would also take `+e` in to consideration.

What I'm thinking is something like
```c++
ModResult OnCheckBan(User* user, Channel* c, const std::string& mask)
{
    if ((mask.length() > 2) && (mask[0] == 'b') && (mask[1] == ':') && !checking)
    {
        Channel* chan = ServerInstance->FindChan(mask.substr(2));
        if (!chan)
            return MOD_RES_PASSTHRU;

        this->checking = true;
        bool matched = chan->IsBanned(user);
        this->checking = false;
        if (matched)
            return MOD_RES_DENY;
    }
    return MOD_RES_PASSTHRU;
}
```

genius3000: I don't recall if I tested such a situation but just going from memory it wouldn't get in an endless loop. All ExtBans run a check on mask.substr(2) (or similar) which will still end up with this module receiving "b:#channel" and it will stop there.  
I'm not sure I want to check exceptions. I'm open to community feedback on this, but personally I see including another channel's banlist as just that, nothing else.

linuxdaemon: `MODE #channel +b U:b:#channel1`
`MODE #channel1 +b U:b:#channel`
After this, it will reach an infinite loop.

genius3000: Hmmm yeah I see your point here. Although, upon testing I'm not seeing any issues, which is strange. Either way, I will rework it for better safety.

sebastiannielsen:
I don't Think a check is neccessary, since actually, anyone, can read the banlist of a "foregin" channel (just run /mode [channel] +b). Thus, nothing prevents a channel owner from copying that banlist or even copying it with a automatic bot that polls the banlist and copies it into own channel, even without this extban.

PR #128, opened 2018-03-17
Add user-specific setting to m_slowmode

linuxdaemon:
This adds the optional `u` and `c` prefix to the slowmode settings, allowing `m_slowmode` to be applied to either the entire channel or on a user-by-user basis.


linuxdaemon:
@SaberUK Done

PR #132, opened 2018-03-26
Add m_blockinvite

### LINKED ISSUE #131: Add Umode +V to disallow receiving /INVITEs
InspIRCd should perhaps have a way that users may prevent receiving invites at all. This could perhaps be accomplished via a usermode, such as +V in the event of for example being mass spammed by invite spam bots. 
The idea of a user mode is so that if a server/network is experiencing ways of spam invites, network staffers can easily make this a (temporary) default for all users immediatly on connect via m_conn_umodes instead of having each user having to type /silence +*!*@* i

Thank you for your time and efforts in advance.

Regards, 

Koragg

genius3000:
Usermode 'V' to block all INVITEs

Resolves #131

### DIFF
--- a/2.0/m_blockinvite.cpp
+++ b/2.0/m_blockinvite.cpp
@@ -1,0 +1,65 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 genius3000 <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Provides usermode 'V' - block all INVITEs */
+/* $ModDepends: core 2.0 */
+
+/* Helpop Lines for the UMODES section
+ * Find: '<helpop key="umodes" value="User Modes'
+ * Place just before the 'W    Receives notif...' line
+ V            Blocks all INVITEs from other users (requires
+              blockinvite extras-module).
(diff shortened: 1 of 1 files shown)


2.0/m_blockinvite.cpp
```diff
@@ -0,0 +1,65 @@
...
+ * Find: '<helpop key="umodes" value="User Modes'
+ * Place just before the 'W    Receives notif...' line
+ V            Blocks all INVITEs from other users (requires
+              blockinvite extras-module).
+ */
+
+#include "inspircd.h"
+
+
+class BlockInvite : public SimpleUserModeHandler
```
SaberUK: Minor nit: you don't need to inherit from SimpleUserModeHandler unless you're changing behaviour in the class. I know 2.0 does this for some stuff but its unnecessary.

```c++
private:
    SimpleUserModeHandler bi;

public:
    ModuleBlockInvite()
        : bi(this, "blockinvite", 'V')
    {
    }

SaberUK:
LGTM other than that one minor nit.

2.0/m_blockinvite.cpp
genius3000: Makes sense, I did initially do like privdeaf with the ModeChange() catch, but decided against needing that.

KoraggKnightWolf:
Dear InspIRCd Developers,

Thank you very much for implementing this module request so quickly. I have tested it and would just like to make two suggestions: 
1. What about informing the person sending the /INVITE that the recipient has umode +V set (or does not accept /INVITE)? Similar to how +D or +T inform the sender perhaps?
2. What about adding in oper_override so that opers can still send out /INVITEs? 

Thank you for your time and effort in advance once more. 

Regards,

Koragg

genius3000:
1) Well understandable in a legit, /invite situation; this is undesirable in a spam defence measure. Supplying a denial message only gives them the ability to change spam tactics.  
2)  I didn't see oper override as necessary, though I rarely use /invite nor need it. It's your module request, so if it fits your use-case.

KoraggKnightWolf:
1. Thank you for pointing that out I indeed overlooked that though I am not sure, do most bots not just execute the same tatctic over and over again like on autopilot? Either way, I overlooked that and it is indeed a valid point you found in there.

2. It might be useful that opers can still /invite users e.g. to #help or such to inform them about things going on or for example in another dedicated network feedback channel like in our case. Would those use cases justify an oper-override?

Thank you very much for taking the time to reply to my requests and to elaborate on my thoughts and also poitning out thought flaws when they are there. Also I hope of course that any network running InspIRCd can benefit from this module.

Regards,

Koragg

KoraggKnightWolf:
How about that oper-override is configurable similar to how it is in the m_noctcp_user module? That way both cases are servced, for networks that wish opers to override +V and those that do not.

PR #133, opened 2018-03-29
Update m_blockinvite as per request

genius3000:
**Added:**
* Oper override using the `users/override-blockinvite` priv  
This seems like a better way than a simple config option, as it lets you configure more specifically which oper class can override.
* Config option to send a reply when the INVITE is blocked: `<blockinvite reply="yes">`

**Fixed:**
* Acting on local or remote source user, due to only checking IS_LOCAL on dest.  
The correct way (necessary now) is to check IS_LOCAL on source and act accordingly.

@KoraggKnightWolf You giving this a test and feedback would be greatly appreciated. I'll keep it a WIP until you're good with it.

### DIFF
--- a/2.0/m_blockinvite.cpp
+++ b/2.0/m_blockinvite.cpp
@@ -20,6 +20,12 @@
 /* $ModAuthorMail: genius3000@g3k.solutions */
 /* $ModDesc: Provides usermode 'V' - block all INVITEs */
 /* $ModDepends: core 2.0 */
+/* $ModConfig: <blockinvite reply="no"> */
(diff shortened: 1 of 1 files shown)


KoraggKnightWolf:
Dear genius3000,

Thank you very much for opimizing the requested module I have tested the functionality and the oper-override when granted works as intended as the message if reply is set to yes. 

All features are working as intended, testing is positive.

I hope that others can benefit from this module as well, at the very least as a preventive measurement against mass invite waves.

Regards,

Koragg

PR #137, opened 2018-04-06
Add m_randomnotice - Random notices at a set interval

### LINKED ISSUE #42: Request m_randomnotice
I would like to request a module that sends local users a random notice every X minutes. For example `<randomnotice delay="5" file="data/randomnotice.db">` so every 5 minutes the local users would receive a notice from the server with a random line from `randomnotice.db`.

genius3000:
A re-work of insp20's vendor module m_randquote.
Modified to send on an interval setting instead of on connect or with a command.
Still reads from a file and keeps the suffix and prefix options.

Resolves #42

### DIFF
--- a/2.0/m_randomnotice.cpp
+++ b/2.0/m_randomnotice.cpp
@@ -1,0 +1,123 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 genius30000 <genius3000@g3k.solutions>
+ *   Copyright (C) 2010 Daniel De Graaf <danieldg@inspircd.org>
+ *   Copyright (C) 2007-2008 Robin Burchell <robin+git@viroteck.net>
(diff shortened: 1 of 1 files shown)


2.0/m_randomnotice.cpp
```diff
@@ -0,0 +1,123 @@
...
+	{
+		OnRehash(NULL);
+		Implementation eventlist[] = { I_OnBackgroundTimer, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+	}
+
+	void OnRehash(User*)
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("randomnotice");
+		std::string n_file = tag->getString("file", "data/randomnotices.db");
```
SaberUK: The database should be in the conf directory and should end in .txt for consistency with other text files like motd.txt and rules.txt.

genius3000: Fair enough, fixed.

PR #138, opened 2018-04-27
Create m_autooper.cpp

sebastiannielsen:
I want to contribute with this module into inspircd-extras. 

What it does:It adds a "opertype" option to connect block, example:
<connect allow="192.168.0.0/16" name="localoperators" timeout="60" flood="40" threshold="1" pingfreq="120" sendq="262144" recvq="8192" localmax="10" globalmax="10" opertype="NetAdmin">

You simple define a connect block extra for each IP/network you want to give oper status, and they gain oper status automatically upon connect. 

If a connect block has no opertype value, user matching this will not get opered up.

The good with this module, is that all parsing and interpretation of IPs/CIDR masks is done in core with well-proven code, which means theres no security risk in this module that would allow an unauthorized user to gain oper status via for example quirks in IPs/CIDR parsing code. (unless the server admin is stupid and writes something like <connect allow="*" opertype="NetAdmin"> but then he has itself to blame for giving everyone Oper status)

### DIFF
--- a/2.0/m_autooper.cpp
+++ b/2.0/m_autooper.cpp
@@ -1,0 +1,52 @@
+/*
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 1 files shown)


2.0/m_autooper.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+#include "inspircd.h"
+
+class ModuleAutoOper : public Module
+{
+public:
+        void init()
+        {
+                Implementation eventlist[] = { I_OnUserConnect };
```
SaberUK: If you're only hooking onto one event then you can just pass it to Attach rather than using an implementation list.

sebastiannielsen:
Done

PR #139, opened 2018-05-03
Create module m_opban.cpp

sebastiannielsen:
Thos module implements a new extban type o:, which prevents the banned person from being admin'ed,  opped, half-opped or voiced.
The extban o: can only be set or unset by someone having +q in the channel, and/or opers.

If the person setting the mode, and/or the target is an oper, the extban check is bypassed. (This to avoid any clash with any oper override modules being present).

Important note: If the OP in question have chanserv OP, be sure to set his access right so he ONLY have auto-OP, not the possibility to OP other's, as ulines also bypass the extban check.

This module is perfect if a channel owner wants to prevent a bad apple from becoming OP, either by social engiinering the present OPs, or aquiring it in some other way, while still retaining the ability for OPs to OP up other good persons.

### DIFF
--- a/2.0/m_opban.cpp
+++ b/2.0/m_opban.cpp
@@ -30,7 +30,7 @@ public:
 	ModResult OnRawMode(User* user, Channel* chan, const char mode, const std::string &param, bool adding, int pcnt)
 	{
 		if (chan && IS_LOCAL(user) && !IS_OPER(user) && !ServerInstance->ULine(user->server)) {
-	    		Membership* transmitter = chan->GetUser(user);
+			Membership* transmitter = chan->GetUser(user);
 			if ((mode == 'b') && (param.length() > 2) && (param[0] == 'o') && (param[1] == ':') && (transmitter->modes.find('q') == std::string::npos))
 			{
 				user->WriteNumeric(482, "%s %s : You must have channel owner access or oper to be able to set or unset extban o", user->nick.c_str(), chan->name.c_str());
--- a/2.0/m_opban.cpp
+++ b/2.0/m_opban.cpp
@@ -1,0 +1,60 @@
+/*
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: Sebastian Nielsen */
+/* $ModAuthorMail: sebastian@sebbe.eu */
+/* $ModDesc: Implements extban +b o: Prevents these persons from aquiring a privilegied position */
+/* $ModDepends: core 2.0 */
+
+#include "inspircd.h"
+class ModuleOPBan : public Module {
+public:
+  void init()
+  {
+    Implementation eventlist[] = { I_OnRawMode, I_On005Numeric };
+    ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+  }
+  
+  Version GetVersion()
+  {
+    return Version("Implements extban +b o: Prevents these persons from aquiring a privilegied position",VF_OPTCOMMON);
+  }
+
+  ModResult OnRawMode(User* user, Channel* chan, const char mode, const std::string &param, bool adding, int pcnt)
+  {
+    if (chan && IS_LOCAL(user) && !IS_OPER(user) && !ServerInstance->ULine(user->server)) {
+      Membership* transmitter = chan->GetUser(user);
+      if ((mode == 'b') && (param.length() > 2) && (param[0] == 'o') && (param[1] == ':') && (transmitter->modes.find('q') == std::string::npos))
+      {
+				user->WriteNumeric(482, "%s %s : You must have channel owner access or oper to be able to set or unset extban o", user->nick.c_str(), chan->name.c_str());
+				return MOD_RES_DENY;
+			}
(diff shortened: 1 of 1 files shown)


2.0/m_opban.cpp
```diff
@@ -0,0 +1,60 @@
...
+				return MOD_RES_DENY;
+			}
+			if (adding && !param.empty())
+			{
+				User *u = ServerInstance->FindNick(param);
+				if (u)
+				{
+					if ((chan->GetExtBanStatus(u, 'o') == MOD_RES_DENY) && !IS_OPER(u) && (mode == 'a' || mode == 'o' || mode == 'h' || mode == 'v'))
+					{
+						user->WriteNumeric(482, "%s %s : %s is banned from having a privilegied position", user->nick.c_str(), chan->name.c_str(), u->nick.c_str());
```
SaberUK:
```diff
- privilegied
+ privileged
```

2.0/m_opban.cpp
```diff
@@ -0,0 +1,60 @@
...
+			{
+				user->WriteNumeric(482, "%s %s : You must have channel owner access or oper to be able to set or unset extban o", user->nick.c_str(), chan->name.c_str());
+				return MOD_RES_DENY;
+			}
+			if (adding && !param.empty())
+			{
+				User *u = ServerInstance->FindNick(param);
+				if (u)
+				{
+					if ((chan->GetExtBanStatus(u, 'o') == MOD_RES_DENY) && !IS_OPER(u) && (mode == 'a' || mode == 'o' || mode == 'h' || mode == 'v'))
```
SaberUK: This will break if anyone remaps those mode characters to something else or if someone adds a prefix mode with a different rank.

You probably want to look the mode up and check if it has a prefix rank associated with it.

2.0/m_opban.cpp
```diff
@@ -0,0 +1,60 @@
...
+		return Version("Implements extban +b o: Prevents these persons from aquiring a privilegied position",VF_OPTCOMMON);
+	}
+
+	ModResult OnRawMode(User* user, Channel* chan, const char mode, const std::string &param, bool adding, int pcnt)
+	{
+		if (chan && IS_LOCAL(user) && !IS_OPER(user) && !ServerInstance->ULine(user->server)) {
+			Membership* transmitter = chan->GetUser(user);
+			if ((mode == 'b') && (param.length() > 2) && (param[0] == 'o') && (param[1] == ':') && (transmitter->modes.find('q') == std::string::npos))
+			{
+				user->WriteNumeric(482, "%s %s : You must have channel owner access or oper to be able to set or unset extban o", user->nick.c_str(), chan->name.c_str());
```
SaberUK:
```diff
- 482
+ ERR_CHANOPRIVSNEEDED
```

2.0/m_opban.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: Sebastian Nielsen */
+/* $ModAuthorMail: sebastian@sebbe.eu */
+/* $ModDesc: Implements extban +b o: Prevents these persons from aquiring a privilegied position */
+/* $ModDepends: core 2.0 */
+
+#include "inspircd.h"
+class ModuleOPBan : public Module {
```
SaberUK: The `{` should be on a new line.

2.0/m_opban.cpp
```diff
@@ -0,0 +1,62 @@
...
+	}
+  
+	Version GetVersion()
+	{
+		return Version("Implements extban +b o: Prevents these persons from aquiring a privileged position",VF_OPTCOMMON);
+	}
+
+	ModResult OnRawMode(User* user, Channel* chan, const char mode, const std::string &param, bool adding, int pcnt)
+	{
+		if (chan && IS_LOCAL(user) && !IS_OPER(user) && !ServerInstance->ULine(user->server)) {
```
SaberUK: It might be better to flatten some of this code structure out with early returns in cases when the majority of the function logic doesn't need to be run. It'd make it a lot easier to read.

2.0/m_opban.cpp
```diff
@@ -0,0 +1,62 @@
...
+			Membership* transmitter = chan->GetUser(user);
+			if ((mode == 'b') && (param.length() > 2) && (param[0] == 'o') && (param[1] == ':') && (transmitter->modes.find('q') == std::string::npos))
```
SaberUK: Rather than hardcoding to q it might be a good idea to have a config option that can take a rank at which users are allowed to remove the mode (see the customprefix ranks in modules.conf.example for the defaults).

2.0/m_opban.cpp
```diff
@@ -12,51 +12,63 @@
...
+		Implementation eventlist[] = { I_OnRawMode, I_On005Numeric, I_OnRehash };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
 	}
-  
 	Version GetVersion()
 	{
 		return Version("Implements extban +b o: Prevents these persons from aquiring a privileged position",VF_OPTCOMMON);
 	}
 
+	void OnRehash(User* user)
```
SaberUK: You should call OnRehash from init so that the config gets loaded on boot.

2.0/m_opban.cpp
```diff
@@ -12,51 +12,63 @@
...
+	{
+		this->RequiredRank = 50000;
+		ConfigTag* conftag = ServerInstance->Config->ConfValue("opban");
+		if (conftag)
```
SaberUK: There's no reason for this if statement. If theres no config tag an empty tag will be returned.

sebastiannielsen: aah, tought it was going to return null, which would result in a runtime error and a crash if I would try to GetInt() on a null value. Fixing.

PR #140, opened 2018-05-05
Add m_badnicks.cpp

sebastiannielsen:
This adds badnicks option to connect class, so you can fine-tune banned nicks in a way not possible with the standard `<badnick>` configuration parameter.

Example config:
```
<connect allow="192.168.0.0/16" name="myoper">
<connect allow="*" name="default" badnicks="*oper* administrator bot*">
```

This would prohibit any nick containing "oper", and any nick exactly matching "administrator", and any nick beginning on "bot", for everyone except those connecting from 192.168.0.0/16

### DIFF
--- a/2.0/m_badnicks.cpp
+++ b/2.0/m_badnicks.cpp
@@ -42,7 +42,7 @@ public:
 		{
 			if (InspIRCd::Match(newnick, badnick))
 			{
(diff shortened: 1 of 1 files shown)


sebastiannielsen:
Or wait, This cannot be used in that context. Must be user.

PR #142, opened 2018-05-08
Add module which prevents a lower rank from removing a listmode entry set by a higher rank

sebastiannielsen:
This module saves the integer rank of a person who sets a listmode item (ban, exception, invex, etc), so if a lower-rank attempts to remove it, it will be rejected. This prevents for example a half-op from reversing a ban set by a op, admin or founder.

*** NOTE TO CHANSERV USERS *** Ensure your operators do not have permission to unban via ChanServ (except for themselves). Also note that any list items set through ChanServ will not be protected, unless ChanServ is present in channel (GUARD ON).

Fixes: https://github.com/inspircd/inspircd/issues/1204

### DIFF
--- a/2.0/m_banprotect.cpp
+++ b/2.0/m_banprotect.cpp
@@ -96,7 +96,8 @@ class ModuleBanprotect : public Module
 			return MOD_RES_PASSTHRU;
 		}
 
-		if (adding) {
+		if (adding)
+		{
 			banp->addrank(mode, param, transmitter->getRank());
 			return MOD_RES_PASSTHRU;
 		}
--- a/2.0/m_banprotect.cpp
+++ b/2.0/m_banprotect.cpp
@@ -24,7 +24,8 @@ class Banprotector
 	void addrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
 	{
 		std::string fullban = ConvToStr(modeparam) + "%" + ConvToStr(banparam);
-		banrank[fullban] = rank;
+		if (!banrank[fullban])
+			banrank[fullban] = rank;
 	}
 
 	bool checkrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
--- a/2.0/m_banprotect.cpp
+++ b/2.0/m_banprotect.cpp
@@ -23,27 +23,51 @@ class Banprotector
 
 	void addrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
 	{
-		if (!banrank[modeparam][banparam])
-			banrank[modeparam][banparam] = rank;
+		std::map<char, std::map<std::string, unsigned int>>::iterator obindex;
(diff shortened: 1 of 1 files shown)


2.0/m_banprotect.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+class Banprotector
+{
+ public:
+	std::map<std::string, unsigned int> banrank;
+	Banprotector() {}
+
+	void addrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
+	{
+		std::string fullban = ConvToStr(modeparam) + "%" + ConvToStr(banparam);
```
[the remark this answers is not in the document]
sebastiannielsen: Fixed. The reason I used a concatenated string instead of map of maps, was that it was both easier and also, I Think it would be more efficient to store a string->rank map instead of a char->string->rank map. It might be a performance increase to use a map of maps as the listmode only needs to be saved once for each banlist, but it depends on how a std::map is saved physically into memory, if the inner maps uses Control codes to delimit the values, it could aswell be same or heavier than using concatenated key strings.

However, as you requested, I have changed into a map of maps instead.

2.0/m_banprotect.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+class Banprotector
+{
+ public:
+	std::map<char, std::map<std::string, unsigned int>> banrank;
+	Banprotector() {}
+
+	void addrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
+	{
+		if (!banrank[modeparam][banparam])
```
SaberUK: You should probably use `.find()` here for each level of the map.

>**Reference to the mapped value of the new element if no element with key key existed.** Otherwise a reference to the mapped value of the existing element whose key is equivalent to key. http://en.cppreference.com/w/cpp/container/map/operator_at

2.0/m_banprotect.cpp
```diff
@@ -0,0 +1,112 @@
...
+			banrank[modeparam][banparam] = rank;
+	}
+
+	bool checkrank(const char& modeparam, const std::string& banparam, const unsigned int& rank)
+	{
+		if (!banrank[modeparam][banparam])
```
SaberUK: Same here as above.

2.0/m_banprotect.cpp
```diff
@@ -0,0 +1,112 @@
...
+		if (banrank[modeparam][banparam] > rank)
+			return false;
+		return true;
+	}
+
+	void delrank(const char& modeparam, const std::string& banparam)
+	{
+		std::map<std::string, unsigned int> innerban = banrank[modeparam];
+		std::map<std::string, unsigned int>::iterator removeindex;
+		removeindex = innerban.find(banparam);
```
SaberUK: See the same as above. Also, innerban might be `banrank.end()`  when using find if this doesn't exist.

PR #145, opened 2018-05-20
Added check to prevent targeting of other oper's

sebastiannielsen:
Added check to prevent /PRETENDUSER on Another oper, which would of course bypass any access restrictions that the executing oper has. Of course, /PRETENDUSER is a pretty serious command and should Only be given to trusted opers, but still, I think it shouldn't be possible to execute this command on Another network operator.

Imagine you dont have rights to /LOADMODULE, but have /PRETENDUSER, but you know the nick of an online oper that has the right.
And you just run /PRETENDUSER [that_oper] LOADMODULE m_restrictchans.so

Can cause pretty serious distruptions in a network.

### DIFF
--- a/2.0/m_pretenduser.cpp
+++ b/2.0/m_pretenduser.cpp
@@ -62,10 +62,18 @@ class CommandPretendUser : public Command
 			return CMD_FAILURE;
 		}
 
-		if (IS_OPER(u))
+		if (IS_OPER(u) && IS_OPER(user))
 		{
-			user->WriteServ("NOTICE %s :*** Cannot target an IRC operator", user->nick.c_str());
-			return CMD_FAILURE;
+			std::string level = u->oper->getConfig("level");
+			long dest_level = atol(level.c_str());
+			level = user->oper->getConfig("level");
+			long source_level = atol(level.c_str());
+
+			if (dest_level > source_level)
+			{
+				user->WriteServ("NOTICE %s :*** Cannot target IRC operator with higher level than yourself", user->nick.c_str());
+				return CMD_FAILURE;
+			}
 		}
 
 		if (IS_LOCAL(u))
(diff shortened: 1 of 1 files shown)


SaberUK:
This is by design. This module is really intended to only be used by someone who has access to everything in the first place.

That said, I guess that allowing opers to become exempt from it would be a good idea. I don't think using `IS_OPER` is a good check for this though. Maybe use `user->HasPrivPermission("users/pretenduser-target-opers")` or something like that.

sebastiannielsen:
Why isn't IS_OPER a good check for that? I can't see any situation where it would be suitable to execute a command as Another oper.

SaberUK:
>Why isn't IS_OPER a good check for that? I can't see any situation where it would be suitable to execute a command as Another oper.

Any reason you might want to run it as normal user is valid for an oper, although now I think about it integrating with m_operlevels might be more appropriate than a privilege.

sebastiannielsen:
I agree about m_operlevels . Could look into that.

2.0/m_pretenduser.cpp
```diff
@@ -62,10 +62,18 @@ class CommandPretendUser : public Command
 			return CMD_FAILURE;
 		}
 
-		if (IS_OPER(u))
+		if (IS_OPER(u) && IS_OPER(user))
 		{
-			user->WriteServ("NOTICE %s :*** Cannot target an IRC operator", user->nick.c_str());
-			return CMD_FAILURE;
+			std::string level = u->oper->getConfig("level");
```
SaberUK: Rather than storing the level in a temporary variable and then calling atoi on its `.c_str()` you should just call ConvToInt here and store directly into dest_level and source_level.

sebastiannielsen: The strange thing is that m_operlevels.so store in a temp variable and uses longs. They are longs so I took the assumption it would be wrong to use integers here.

SaberUK: Yeah, operlevels does that in 2.0 but it has been changed for 3.0.

PR #146, opened 2018-05-25
Add module for restricting channel modes to specific prefix ranks

sebastiannielsen:
This module allows restricting modes to specific prefix ranks.

For example:
```
<moderestrict mode="b" rank="30000"> = half-op cannot ban/unban
<moderestrict mode="v" rank="30000"> = half-op cannot voice/unvoice
<moderestrict mode="t" rank="50000"> = MLOCK functionality
<moderestrict mode="o" rank="40000"> = Only channel admin and above can op/deop
<moderestrict mode="w" rank="40000"> = Only channel admin and above can edit autoop list
```

Note that this module can NOT be used to lower the requirement for a channel mode already present in inspircd, only make it stricter.

In combination with: m_autoop, m_chanprotect, m_halfop, this can be used to create a services-less server that have roughtly the functionality of one with services, but uses hostnames/IP-adresses to identify users instead.

### DIFF
--- a/2.0/m_moderestrict.cpp
+++ b/2.0/m_moderestrict.cpp
@@ -61,7 +61,7 @@ class ModuleModeRestrict : public Module
 
 		if (transmitter->getRank() < mc->second)
 		{
-			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, "%s %s : You do not have sufficient privileges to set or unset mode %c", user->nick.c_str(), chan->name.c_str(), mode);
+			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, "%s %s :You do not have sufficient privileges to set or unset mode %c", user->nick.c_str(), chan->name.c_str(), mode);
 			return MOD_RES_DENY;
 		}
 		return MOD_RES_PASSTHRU;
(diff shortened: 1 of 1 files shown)


2.0/m_moderestrict.cpp
```diff
@@ -0,0 +1,72 @@
...
+	ModResult OnRawMode(User* user, Channel* chan, const char mode, const std::string &param, bool adding, int pcnt)
+	{
+		if (!chan || !IS_LOCAL(user) || IS_OPER(user) || ServerInstance->ULine(user->server))
+			return MOD_RES_PASSTHRU;
+
+		Membership* transmitter = chan->GetUser(user);
+		if (!transmitter)
+			return MOD_RES_PASSTHRU;
+
+		std::map<char, unsigned int>::iterator mc;
```
SaberUK: You can assign to this at the same time as creating it.

2.0/m_moderestrict.cpp
```diff
@@ -0,0 +1,70 @@
...
+		if (!transmitter)
+			return MOD_RES_PASSTHRU;
+
+		std::map<char, unsigned int>::iterator mc = requiredrank.find(mode);
+		if (mc == requiredrank.end())
+			return MOD_RES_PASSTHRU;
+
+		if (transmitter->getRank() < mc->second)
+		{
+			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, "%s %s : You do not have sufficient privileges to set or unset mode %c", user->nick.c_str(), chan->name.c_str(), mode);
```
SaberUK: You shouldn't have a space beore the message in the <trailing> parameter.

PR #149, opened 2018-06-25
Add a module to limit SASL connect classes by account(s).

genius3000:
Allow "SASL gates" (connect classes for authenticated users) to be limited to specified account(s).

Idea/request by @Robby-. 

### Testing
* Successfully denies a non-matching account
* Successfully doesn't deny a matching account (other parameters can still deny)
* Disabling `requireaccount` disables this
* With `m_services_account ` unloaded, connections are denied to any class with this enabled

### DIFF
--- a/2.0/m_conn_accounts.cpp
+++ b/2.0/m_conn_accounts.cpp
@@ -1,0 +1,79 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 Matt Schatz <genius3000@g3k.solutions>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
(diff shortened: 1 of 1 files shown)


2.0/m_conn_accounts.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+
+class ModuleConnAccounts : public Module
+{
+	const AccountExtItem* accountname;
+
+ public:
+	ModuleConnAccounts()
+	{
+		accountname = GetAccountExtItem();
```
SaberUK: What happens to this if `m_services_account` unloads?

genius3000: Let me run another test or two on variations regarding this, but I did unload m_services_account and connect a client to match a class to this module and it denied it, as expected. I'll make sure it handles loading without okay too.

genius3000: Alright so no crashes any which way I tried but one scenario isn't great:
- starting Insp with this (or loading this while running) but not m_services_account: denies anyone to a class with `accounts`
   - loading m_services_account after: requires reload of this to get the ExtItem
- unloading m_services_account: same as starting without m_services_account
- reloading m_services_account: continues to work. 

I could go two routes to better this:
1. Remove the const and catch OnLoadModule to update the ExtItem.
2. Switch to getting the ExtItem each time within OnSetConnectClass, which is similar to how m_require_auth works.  

Easier and safer way would be #2 but increased number of ExtItem gets. Any suggested preference?

genius3000:
Updated with the better/safer handling of AccountExtItem.

PR #150, opened 2018-07-08
Backport m_showfile from a 3.0 vendor module to a 2.0 extras

genius3000:
Did this up after @KoraggKnightWolf was looking for this functionality in 2.0. For reference, a [diff from 3.0](https://gist.githubusercontent.com/genius3000/4e05505a4847c0fb818889485383c124/raw/43f6a326b71cd4d81329ef334128d53fc712cec9/m_showfile.diff) to this.

Tested on my test net with each `method`, empty values, and invalid numerics. All that and proper configurations, including formatting codes in the text files work as expected.

### DIFF
--- a/2.0/m_showfile.cpp
+++ b/2.0/m_showfile.cpp
@@ -1,0 +1,214 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 Matt Schatz <genius3000@g3k.solutions>
+ *   Copyright (C) 2013 Attila Molnar <attilamolnar@hush.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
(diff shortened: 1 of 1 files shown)


2.0/m_showfile.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+			// This is our command, make sure we don't have the same entry twice
+			sfcmd = static_cast<CommandShowFile*>(handler);
+			if (std::find(newcmds.begin(), newcmds.end(), sfcmd) != newcmds.end())
+				throw ModuleException("Command " + cmdname + " is already used in a <showfile> tag");
+		}
+		else
+		{
+			// Command doesn't exist, create it
+			sfcmd = new CommandShowFile(this, cmdname);
```
SaberUK: If AddService fails here then the memory pointed to by sfcmd will be leaked.

genius3000:
Just a reminder, discussed the review comment on IRC and AddService only fails if the command exists, which is checked for already.

PR #156, opened 2018-07-30
m_joinpartspam: Bug-fix and comment update

genius3000:
* Fix a bug/workaround of using /CYCLE to cycle the channel countless times:
  * Move most of the OnUserPreJoin code to a separate function.
  * Use this new function for both OnUserPreJoin and OnPreCommand (CYCLE).
  * CYCLE will be denied if the user is or would be blocked when joining. The NOTICE sent is similar to m_cycle's "you are banned" denial message. CYCLE will not get redirected, same as how a ban redirect won't affect it.

* Update a code comment that was missed in the rewrite from banning to blocking and later with the addition of a /INVITE block removal function.

### DIFF
--- a/2.0/m_joinpartspam.cpp
+++ b/2.0/m_joinpartspam.cpp
@@ -330,7 +330,7 @@ class ModuleJoinPartSpam : public Module
 		OnRehash(NULL);
 		ServiceProvider* servicelist[] = { &ext, &jps };
 		ServerInstance->Modules->AddServices(servicelist, sizeof(servicelist)/sizeof(ServiceProvider*));
-		Implementation eventlist[] = { I_OnRehash, I_OnUserPreJoin, I_OnUserJoin, I_OnUserInvite };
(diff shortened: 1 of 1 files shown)


2.0/m_joinpartspam.cpp
```diff
@@ -376,9 +371,39 @@ class ModuleJoinPartSpam : public Module
...
+	{
+		if (result == CMD_FAILURE || command != "CYCLE" || IS_OPER(user))
+			return;
+
+		Channel* chan = ServerInstance->FindChan(parameters[0]);
+		if (!chan || !chan->IsModeSet(jps.GetModeChar()))
+			return;
+
+		bool kick = BlockJoin(user, chan);
+		if (kick)
```
SaberUK: Rather than kicking could you not check in OnPreCommand as to whether the user will be blocked?

genius3000:
Forgot to comment before, this has been updated to prevent the use of CYCLE instead of kicking after.

PR #159, opened 2018-08-14
Add module for snomasks p and P (User PMs)

sebastiannielsen:
I got a request for creating this module from a server owner on Another network via email, since this owner had large problems with abuse over PMs that fly under the radar.
This module allows server administrators (and opers) to read PMs through Snomasks p and P. (p = local PMs, P = remote PMs).
This allows server administrators to enforce network usage rules across PMs.
Note that on larger networks this module might get spammy, why I recommend instead redirecting the output to a channel like #userpm using m_chanlog.so, and don't forget +O on that channel.

Config:
`<userpm ignoreuline="yes/no" ignoreoper="yes/no" expandnicks="0/1/2">`

ignoreuline="yes"/"no": This will make the module ignore PMs to/from u-lined servers if set. This config is HIGHLY RECOMMENDED to set to "yes" to avoid NickServ/ChanServ/OperServ passwords from ending up in the wrong places. However, for some networks it might be useful to set this to "no" to make it possible to log excessive login attempts to NickServ/ChanServ/OperServ and use a bot to issue Z-lines/G-lines/etc.

ignoreoper="yes"/"no": This will make the module ignore PMs to/from opers. In some cases it can be a good idea to set it to "yes" to avoid PMs to/from Opers from ending up as server notices, but in some cases it can be useful to see other oper's PMs (for example if an oper does an decision in favor for a user, for example "Yes you can bring your own bot here", it could be useful for a network at a large to be able to see this decision and prevent other oper's from intervening thinking the user did not have permission)

expandnicks="0/1/2":
This config allows you to set how nicks are displayed in the server notices. If you set this to 0, only the nick is shown, for example:
`USERPM: <Source>,<Target>: Content of PM`

If you set it to 1, it will display the nicks in expanded nick!ident@host format, but will leave host cloaked, if you have the cloak module activated, for example:
`USERPM: <Source!Source@network-gs7.9vf.8gbi23.IP>,<Target!Target@network-h8j.2dv.2nbjkl.IP>: Content of PM`

If you set it to 2, it will display it in uncloaked format, for example:
`USERPM: <Source!Source@user1.someisp.com>,<Target!Target@user2.anotherisp.com>: Content of PM`

The module displays everything in a easy-to-parseable format so its also possible to create bots that act on PM content.
To disable the possibility to use OTR encryption, use a censor/filter module to disable the string "?OTR" to be sent.

Note that this module can be a Little controversial in the IRC World, why I recommend to be responsible when using this module on your server, and not allowing every oper to use this module. More about restricting usage to certain opers later down.

To restrict which opers that are able to read PMs through snomasks, I recommend disabling mode "s" in server config (so no opers can use SNOMASKS), and instead provide SNOMASK functionality through separate +O channels using m_chanlog.so, where you can through channel config restrict which opers have access (for example via bans, exceptions, invites and channel keys), which allows diverse access condition for each snomask.

### DIFF
--- a/2.0/m_userpm.cpp
+++ b/2.0/m_userpm.cpp
@@ -1,0 +1,85 @@
+/* This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+
+/* $ModAuthor: Sebastian Nielsen */
+/* $ModAuthorMail: sebastian@sebbe.eu */
+/* $ModDesc: Provides snomasks p and P to which notices about PMs are sent */
+/* $ModDepends: core 2.0 */
+/* $ModConfig: Optional: <userpm ignoreuline="no" ignoreoper="no" expandnicks="0/1/2">, 0 = nick only, 1 = full + cloak, 2 = full */
+
+class ModuleUserPM : public Module
+{
+ private:
+ 	bool IgnoreUline;
+	bool IgnoreOper;
+	int ExpandNicks;
+ public:
+	void init()
+	{
+		ServerInstance->SNO->EnableSnomask('p', "USERPM");
+		Implementation eventlist[] = { I_OnUserMessage, I_OnUserNotice, I_OnRehash };
+		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+		OnRehash(NULL);
+	}
+
+	Version GetVersion()
+	{
+		return Version("Provides snomasks p and P to which notices about PMs are sent");
+	}
+
+	void OnRehash(User *user)
+	{
+		ConfigTag* conftag = ServerInstance->Config->ConfValue("userpm");
+		this->IgnoreUline = conftag->getBool("ignoreuline", false);
+		this->IgnoreOper = conftag->getBool("ignoreoper", false);
+		this->ExpandNicks = conftag->getInt("expandnicks", 0);
+	}
+
+	void OnUserMessage(User *user, void *dest, int targettype, const std::string &pm, char status, const CUList &except)
+	{
+		if ( (targettype != TYPE_USER) || !IS_LOCAL(user) )
+		{
+			return;
+		}
+		User *duser = (User *)dest;
+		if (this->IgnoreUline && ( ServerInstance->ULine(user->server) || ServerInstance->ULine(duser->server) ))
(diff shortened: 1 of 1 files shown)


SaberUK:
I'm sorry but i'm not going to merge modules that allow operators to silently spy on private messages between users without their consent. This is not only unethical but it is also *illegal* in [a significant chunk of the world](https://en.wikipedia.org/wiki/Telephone_recording_laws).

Users should have to explicitly opt-in to any kind of monitoring of private messages (e.g. a user mode) and should be informed in a very noticeable way (e.g. sending the user a `PRIVMSG` informing them that they have enabled PM redirection).

sebastiannielsen:
I wonder what the problem is? Following laws is the responsibility of the server administrator.
I could add a auto-notice feature that is default on, is that OK, that will inform on connect?

There is different laws in different countries.
Could add "lawmode" parameter:
lawmode = 0 - silent
lawmode = 1 - notice on connect (default)
lawmode = 2 - consent required for ONE PARTY
lawmode = 3 - consent required for BOTH PARTY

About wiretapping laws, there isn't any Technical difference between a channel message and user PM, only difference is that the former has a # prefixed to the target. Thus m_override.so would also be "illegal" then, if opers spy in channels?

For example in sweden, wiretapping laws only apply to public networks, which are defined as networks that are in possession of a licensed ISP. Thus someone operating a server is free to monitor all and any traffic passing that server (unless that person have a ISP license, then its not allowed, and ISP license is required to be able to sell internet services, opposed to offering them to free).

The idea is that opers that have problems with network abuse over PMs should be able to combat it.

SISheogorath:
No, just no.

Even when the purposed case is maybe legitimate use, this is highly abusable. We already removed m_invisible for that reason (as well as tons of bugs) and I'm sure we don't want to reintroduce something that invasive.

> About wiretapping laws, there isn't any Technical difference between a channel message and user PM, only difference is that the former has a # prefixed to the target. Thus m_override.so would also be "illegal" then, if opers spy in channels?

The important difference is: In channels you see it when someone joins. Means: You see :eyes:  the person reading. PMs are PMs because they should be private and we have enough ways to block PMs either as oper or as user. So I really don't see any need to wiretap user's conversations even in case of abuse. You can create organisational structures for that.

As @SaberUK said, adding a mode to forward things to opers, okay, that's explicit opt-in and works around "faked logs" but wiretap these things by default is a real no, no.

SaberUK:
>I wonder what the problem is? Following laws is the responsibility of the server administrator.

And IRC server administrators have a history of not following laws or caring about ethics.

>Could add "lawmode" parameter:

You can do that if you want but I am still not merging something that allows messages to be silently snooped on without consent regardless of configuration.

>For example in sweden, wiretapping laws only apply to public networks

There's a difference between ethics and laws. Just because some country allows it doesn't mean its not a massive privacy violation to be snooping on peoples private messages without consent.

>The idea is that opers that have problems with network abuse over PMs should be able to combat it.

If someone is receiving PM abuse they can explicitly consent to having their PMs sent to opers.

PR #164, opened 2018-09-03
Backport user mode of m_sslmodes from 3.0

### LINKED ISSUE #154: Backport umode +z from 3.0 vendor m_sslmodes to 2.0 extras
Dear InspIRCd developers,

I hereby would wish to request that the usermode +z from the m_sslmodes module of InspIRCd 3.0 may be backported as an extras module for InspIRCd 2.0. It would on the one hand solve the mentioned backwards compatbility issues mentioned in the user mode option in 3.0 and allow users to restrict queries to only those on an SSL connection. 
The benefit would be a stronger sense of securty and privacy via SSL only queries and could also be used as a "gentler" option to keep out spambots which mass pm as thse (often) do not use SSL which might be less restrictive than +R and +g.

Thank You very much for Your time and effort in advance.

Regards,

Koragg

genius3000:
The user mode 'z' was added to the vendor m_sslmodes for 3.0 to allow SSL/Secure only queries and user notices. Backport just the user mode functionality with the same mode name for 2.0.

Tested on single server and multiple servers, blocks when it should.

Resolves #154

### DIFF
--- a/2.0/m_sslmodeuser.cpp
+++ b/2.0/m_sslmodeuser.cpp
@@ -1,0 +1,146 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 Matt Schatz <genius3000@g3k.solutions>
+ *   Copyright (C) 2013 Shawn Smith <shawn@inspircd.org>
+ *   Copyright (C) 2009 Daniel De Graaf <danieldg@inspircd.org>
+ *   Copyright (C) 2007 Robin Burchell <robin+git@viroteck.net>
+ *   Copyright (C) 2007 Dennis Friis <peavey@inspircd.org>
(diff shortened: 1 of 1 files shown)


2.0/m_sslmodeuser.cpp
```diff
@@ -0,0 +1,144 @@
...
+/* $ModAuthorMail: genius3000@g3k.solutions */
+/* $ModDesc: Provides user mode +z to allow for Secure/SSL only queries and user notices */
+/* $ModDepends: core 2.0 */
+
+#include "inspircd.h"
+#include "ssl.h"
+
+
+// Common function throughout this module
+bool IsSSLUser(Module* mod, User* user)
```
SaberUK: This should either be in an anonymous namespace or should be static.

genius3000: Yep, my bad...namespace it is.

genius3000:
Updated

Issue #155, opened 2018-07-27
m_conn_invite


stargatecmd2009:
Hi,
Are we able to have a module that auto invites on connect instead of force joining like how m_conn_join does. While possibly still respecting any bans etc that are set in the channel.
With extra config/tag settings that allow it to be enabled/disabled on certain servers and connect classes.

`<autoinvite channel="#one,#two,#three">`
`<autoinviteexempt type="server" pattern="*.relay">`

damiankw:
Hey @stargatecmd2009, do you want to give my PR a try and see if it works? I don't think you need to add the `<autoinviteexempt>` tag, if you don't want to have a specific server auto invite, just don't add the configuration into your inspircd.conf file, you will probably still have to load the module so it's compliant with the rest of the network.

Please note, I haven't actually tested this on a multi-server network yet, I would recommend testing on a dev network before putting it into production.

SaberUK:
Implemented in #168.

PR #168, opened 2018-11-03
m_conn_invite

damiankw:
Adds ability to automatically invite users connecting to the network to specific channels. First version which doesn't have any options. Part of #155.


SaberUK:

PR #174, opened 2018-12-24
Add m_sasl_servercheck - Fail SASL auth if the target is unavailable

genius3000:
m_sasl can be configured with a `target` value to direct the ENCAP messages to only that server. Some clients don't handle a SASL no reply very well and just timeout. This sends back a SASL Fail numeric if the configured target is unavailable. This is just a "band-aid" for 2.0 (for those that wish to use it). Insp 3.0 handles this correctly with the SASL CAP being unavailable if the target server is offline.

### DIFF
--- a/2.0/m_sasl_servercheck.cpp
+++ b/2.0/m_sasl_servercheck.cpp
@@ -1,0 +1,79 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018 Matt Schatz <genius3000@g3k.solutions>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* $ModAuthor: genius3000 */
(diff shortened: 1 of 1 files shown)


2.0/m_sasl_servercheck.cpp
```diff
@@ -0,0 +1,79 @@
...
+	void OnRehash(User*)
+	{
+		reason = ServerInstance->Config->ConfValue("saslservercheck")->getString("reason", "SASL is currently unavailable.");
+		target = ServerInstance->Config->ConfValue("sasl")->getString("target");
+
+		if (target.empty() || target == "*")
+			throw ModuleException("This module is useless without setting the <sasl target=\"services.mynetwork.com\"> value");
+	}
+
+	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser* user, bool validated, const std::string &original_line)
```
SaberUK:
```suggestion
	ModResult OnPreCommand(std::string& command, std::vector<std::string>&, LocalUser* user, bool validated, const std::string&)
```

2.0/m_sasl_servercheck.cpp
```diff
@@ -0,0 +1,79 @@
...
+	{
+		if (command != "AUTHENTICATE" || (target.empty() || target == "*"))
```
SaberUK: You should add a check here for the command not being validated?

2.0/m_sasl_servercheck.cpp
```diff
@@ -0,0 +1,79 @@
...
+			return MOD_RES_PASSTHRU;
+
+		ProtoServerList servers;
+		ServerInstance->PI->GetServerList(servers);
+		if (servers.empty())
+			return MOD_RES_PASSTHRU;
+
+		for (std::list<ProtoServer>::const_iterator i = servers.begin(); i != servers.end(); ++i)
```
SaberUK:
```suggestion
		for (ProtoServerList::const_iterator i = servers.begin(); i != servers.end(); ++i)
```

genius3000:
Thanks for the review; concerns addressed.

PR #175, opened 2018-12-24
Update to m_xlinetools - bug fix and improvements

genius3000:
1. Add `XCOUNT`
   * Simply just counts the number of matched lines.
   * Re-order and rename some functions.
   * Documentation update
2. Improve the matching logic
   * Cut back repetitive calls to Match() and makes the code a little cleaner.
3. Bug-fix: `XCOPY` and `XREMOVE` could be used by any oper regardless of command permissions
   * Check for appropriate command permission before allowing.
   * Switched to local only command handlers to accommodate this.
   * Moved common functions inside an anonymous namespace.
4. Add `setby` as an alias for `source` as it sounds more familiar.

### DIFF
--- a/2.0/m_xlinetools.cpp
+++ b/2.0/m_xlinetools.cpp
@@ -275,6 +275,8 @@ class CommandXBase : public SplitCommand
 	{
 		total += xlines->size();
 		LookupIter safei;
+		bool negate;
+		bool match;
 
 		for (LookupIter i = xlines->begin(); i != xlines->end(); )
 		{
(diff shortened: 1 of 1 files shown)


Robby-:
This is nitpicking but would it be possible to change `Line` to `line` on the following line:
https://github.com/inspircd/inspircd-extras/blob/2e729de932137ca89b89093112bbb6b985110dc8/2.0/m_xlinetools.cpp#L233

and `on` to `for` on the following line:
https://github.com/inspircd/inspircd-extras/blob/2e729de932137ca89b89093112bbb6b985110dc8/2.0/m_xlinetools.cpp#L588

This, so that it becomes consistent with how inspircd itself already words the server notice messages about adding xlines. Although, inspircd itself is again in turn breaking consistency with `Line` in the xline expiry messages because it uses a capital `L` in those messages... sigh. :)

genius3000:
Added a commit to fix some text inconsistencies compared to core (at least to what appears intended) as pointed out by @Robby-.

PR #185, opened 2019-02-12
Add m_kill_idle to disconnect idle users

linuxdaemon:
This allows configuration of multiple `idleprofile` tags which may be applied to a user via `<connect:idleprofile>`.

### DIFF
--- a/3.0/m_kill_idle.cpp
+++ b/3.0/m_kill_idle.cpp
@@ -1,3 +1,21 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2019 linuxdaemon <linuxdaemonirc@gmail.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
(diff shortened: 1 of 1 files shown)


KoraggKnightWolf:
Interesting idea as other IRCd's also have this feature, it is just that the away="" part is not explained anywhere, one would have to guess by the options in the code (no annotations there either etc). Also a port for 2.0 would be very much appreciated if possible. Other than that, very thorough options and per connect class is also good (e.g. so localhost is never hit when it shouldn't be).

Regards,

Koragg

3.0/m_kill_idle.cpp
```diff
@@ -0,0 +1,176 @@
+/// $ModAuthor: linuxdaemon
```
SaberUK: You're missing the copyright header.

3.0/m_kill_idle.cpp
```diff
@@ -0,0 +1,176 @@
...
+		AWAY_IGNORE,
+		AWAY_ONLY,
+		AWAY_NONE
+	} away;
+
+	/** Does the user need to be in no channels to match */
+	bool nochans;
+
+	/** Should we ignore opers */
+	bool ignoreopers;
```
SaberUK: Suggestion: make this a privilege instead.

PR #186, opened 2019-02-12
Add m_shed_users to quit users before maintenance

linuxdaemon:
This allows slow disconnects of users when the server is in maintenance
mode, optionally shutting the server down when it is empty


Robby-:
Just in case it is needed, it would be nice to have a way to cancel the shedding.
Currently, unloading the module is the only way.

PR #192, opened 2019-02-22
Refactor user picking logic in shed_users

linuxdaemon:
Sorting the entire local list on every tick is not needed when only quitting one user at a time, so just find the max idle time instead.


SaberUK:
:+1:

PR #194, opened 2019-03-08
m_joinpartspam: Fix prioritization bug.

### LINKED ISSUE #191: m_blockinvite does NOT block invites to +A (allowinvite) chans
Description

When a user has user mode +V set and is invited by any user to a +A channel the user mode +V does NOT block the invite.

Steps to reproduce the issue: 

1. Set a user connection to +V (/mode $nick +V)
2. Have another non oper (or one who may not bypass the mode via users/override-blockinvite) set +A on a channel (mode $chan +A)
3. Have the user from step 2. /INVITE the user from step 1. to the channel created and set +A in step 2.
4. Observe on the user from step 1. that despite being invited by a regular user (or a non +V exempted oper) the invite is NOT blocked, but received as if +V were not set.

Describe the results you received:

The /INVITE was received as if umode +V were not set despite coming from a non opered user (or non umode +V exempted oper)

Describe the results you expected:

The /INVITE should have been blocked due to umode +V as it was sent by a regular non oper (or non umode +V exempted oper)

Additional information you deem important (e.g. issue happens only occasionally):

This issue requires the channel mode +A/the m_allowinvite.so module to be reproduced (shipped with InspIRCd by default).

Output of ./bin/inspircd --version:

InspIRCd-2.0.27 r0

genius3000:
Specifically, m_allowinvite returns ALLOW so we need to run first to be able to return DENY.

Resolves #191.

### DIFF
--- a/2.0/m_blockinvite.cpp
+++ b/2.0/m_blockinvite.cpp
@@ -83,6 +83,13 @@ class ModuleBlockInvite : public Module
 		return MOD_RES_DENY;
 	}
 
+	void Prioritize()
+	{
+		// Go before m_allowinvite as it returns ALLOW.
+		Module* allowinvite = ServerInstance->Modules->Find("m_allowinvite.so");
+		ServerInstance->Modules->SetPriority(this, I_OnUserPreInvite, PRIORITY_BEFORE, allowinvite);
+	}
+
 	Version GetVersion()
 	{
 		return Version("Provides usermode +" + ConvToStr(bi.GetModeChar()) + " to block all INVITEs", VF_OPTCOMMON);


SaberUK:
is there any reason this module needs to be first? Isn't being prioritised in front of allowinvite enough?

genius3000:
Likely not, so yeah just in front of allowinvite should be fine. Is it generally better to keep prioritization specific when possible? Only a few other modules use this hook so first seemed good for a deny rule.

SaberUK:
> Is it generally better to keep prioritization specific when possible?

Yes, because if two modules hook to front the prioritiser will get into a loop until it eventually gives up

PR #196, opened 2019-03-10
Port all of my extras to 3.0.

genius3000:
Port all of my extras to 3.0.
Reviews and testing appreciated.

### Modules List
- [x] m_blockinvite
- [x] m_checkbans
- [x] m_conn_accounts
- [x] m_conn_matchident
- [x] m_conn_require
- [x] m_conn_strictsasl
- [x] m_conn_vhost (minor code improvement)
- [x] m_extbanbanlist
- [x] m_extbanregex
- [x] m_joinpartspam
- [x] m_nocreate
- [x] m_randomnotice
- [x] m_restrictmsg_duration
- [x] m_xlinetools

#### List of my remaining extras not needing a port:
- m_classban: Exists in 3.0.
- m_sasl_servercheck: Handled properly in 3.0.
- m_showfile: Exists in 3.0.
- m_sslmodeuser: Exists in 3.0.

### DIFF
--- a/3.0/m_blockinvite.cpp
+++ b/3.0/m_blockinvite.cpp
@@ -23,7 +23,7 @@
 /// $ModDesc: Provides usermode 'V' - block all INVITEs
 /* Set whether to reply with a blocked message to the source (inviter)
  * with the 'reply' config option. Defaults to no.
- * To allow oper overriding, add the privilege 'users/override-blockinvite'
+ * To allow oper overriding, add the privilege 'users/blockinvite-override'
  * to your preferred oper class.
  */
 
@@ -31,7 +31,7 @@
  * Find: '<helpop key="umodes" value="User Modes'
  * Place just before the 'W    Receives notif...' line
  V            Blocks all INVITEs from other users (requires
-              blockinvite extras-module).
+              the blockinvite extras-module).
  */
 
 
@@ -73,7 +73,7 @@ class ModuleBlockInvite : public Module
 		if (!IS_LOCAL(source) || !dest->IsModeSet(bi.GetModeChar()))
 			return MOD_RES_PASSTHRU;
 
-		if (source->HasPrivPermission("users/override-blockinvite"))
+		if (source->HasPrivPermission("users/blockinvite-override"))
 			return MOD_RES_PASSTHRU;
 
 		if (reply)
--- a/3.0/m_blockinvite.cpp
+++ b/3.0/m_blockinvite.cpp
@@ -1,0 +1,94 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2018-2019 Matt Schatz <genius3000@g3k.solutions>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: genius3000
+/// $ModAuthorMail: genius3000@g3k.solutions
+/// $ModConfig: <blockinvite reply="no">
+/// $ModDepends: core 3.0
+/// $ModDesc: Provides usermode 'V' - block all INVITEs
+/* Set whether to reply with a blocked message to the source (inviter)
+ * with the 'reply' config option. Defaults to no.
+ * To allow oper overriding, add the privilege 'users/override-blockinvite'
+ * to your preferred oper class.
+ */
+
+/* Helpop Lines for the UMODES section
+ * Find: '<helpop key="umodes" value="User Modes'
+ * Place just before the 'W    Receives notif...' line
+ V            Blocks all INVITEs from other users (requires
+              blockinvite extras-module).
+ */
+
+
+#include "inspircd.h"
+
+enum
+{
+	// From UnrealIRCd (channel mode, but same concept)
+	ERR_NOINVITE = 518
+};
+
+class ModuleBlockInvite : public Module
+{
+ private:
+	SimpleUserModeHandler bi;
+	bool reply;
+
+ public:
+	ModuleBlockInvite()
+		: bi(this, "blockinvite", 'V')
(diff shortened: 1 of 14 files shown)


3.0/m_blockinvite.cpp
```diff
@@ -0,0 +1,91 @@
...
+	ModResult OnUserPreInvite(User* source, User* dest, Channel*, time_t) CXX11_OVERRIDE
+	{
+		if (!IS_LOCAL(source) || !dest->IsModeSet(bi.GetModeChar()))
+			return MOD_RES_PASSTHRU;
+
+		if (source->HasPrivPermission("users/override-blockinvite"))
+			return MOD_RES_PASSTHRU;
+
+		if (reply)
+			source->WriteNumeric(ERR_NOINVITE, "%s :Can't INVITE %s, they have +%c set", source->nick.c_str(), dest->nick.c_str(), bi.GetModeChar());
```
SaberUK: `WriteNumeric()` no longer needs the target nick parameter to be passed as the third parameter.

Also, `WriteNumeric()` no longer uses `snprintf()`. Use InspIRCd::Format if you want to format something.

3.0/m_conn_strictsasl.cpp
```diff
@@ -0,0 +1,92 @@
...
+		const AccountExtItem* accountname = GetAccountExtItem();
+		if (!accountname)
+			return MOD_RES_PASSTHRU;
+
+		// Let them through if they have an account
+		std::string* account = accountname->get(user);
+		if (account && !account->empty())
+			return MOD_RES_PASSTHRU;
+
+		ServerInstance->Logs->Log("m_conn_strictsasl", LOG_DEBUG, "Failed SASL auth from: %s (%s) [%s]",
```
SaberUK: You can use `MODNAME` instead of "m_conn_strictsasl" here.

3.0/m_xlinetools.cpp
```diff
@@ -0,0 +1,602 @@
...
+		const std::string mexpires("-expires=");
+
+		for (std::vector<std::string>::const_iterator p = params.begin(); p != params.end(); ++p)
+		{
+			const std::string param = *p;
+
+			if (param.find(mtype) != std::string::npos)
+			{
+				argreason = false;
+				const std::string& val(param.substr(mtype.length()));
```
SaberUK: I'm not sure if this was in the original but this (and below) should be a `std::string` not a `std::string&`.

3.0/m_xlinetools.cpp
```diff
@@ -0,0 +1,602 @@
...
+					settime.c_str(), duration.c_str(), expires.c_str(), reason.c_str()));
+			}
+
+			i = safei;
+		}
+	}
+
+	bool HandleCmd(LocalUser* user, const Criteria& args, Command* cmd)
+	{
+		bool count = (cmd->name == "XCOUNT" ? true : false);
```
SaberUK: Using the ternary operator here is unnecessary as the == operator returns a boolean value you can dirctly assign to count/etc.

3.0/m_xlinetools.cpp
```diff
@@ -0,0 +1,602 @@
...
+	}
+
+ public:
+	CommandXBase(Module* Creator, const std::string& cmdname) : SplitCommand(Creator, cmdname, 1)
+	{
+		syntax = "-type=<type|*> -mask=[!]<> -reason=[!]<> -source=[!]<> -set=[-]<time> -duration=[-+]<time> -expires=[+]<time>";
+		flags_needed = 'o';
+	}
+
+	CmdResult HandleLocal(LocalUser* user, const Params& parameters)
```
SaberUK: This should be `CXX11_OVERRIDE`.

3.0/m_xlinetools.cpp
```diff
@@ -0,0 +1,602 @@
...
+class CommandXCopy : public SplitCommand
+{
+ public:
+	CommandXCopy(Module* Creator) : SplitCommand(Creator, "XCOPY", 3)
+	{
+		syntax = "<X-line type> <old mask> <new mask> [-duration=<> -reason=<>]";
+		flags_needed = 'o';
+	}
+
+	CmdResult HandleLocal(LocalUser* user, const Params& parameters)
```
SaberUK: This should be `CXX11_OVERRIDE`.

3.0/m_restrictmsg_duration.cpp
```diff
@@ -0,0 +1,146 @@
...
+		{
+			if (target.type == MessageTarget::TYPE_USER)
+			{
+				src->WriteNumeric(ERR_CANTSENDTOUSER, target.Get<User>()->nick,
+					InspIRCd::Format("You cannot send messages within the first %lu seconds of connecting.",
+					duration));
+			}
+			else
+			{
+				src->WriteNumeric(ERR_CANNOTSENDTOCHAN, target.Get<Channel>()->name,
```
SaberUK: The target might be `TYPE_SERVER` here. You should update your else statement to explicitly check for TYPE_CHANNEL.

3.0/m_restrictmsg_duration.cpp
```diff
@@ -0,0 +1,146 @@
...
+			return MOD_RES_PASSTHRU;
+
+		// Source is registered (and identified) exemption
+		if (exemptregistered)
+		{
+			const AccountExtItem* accountname = GetAccountExtItem();
+			if (accountname)
+			{
+				const std::string* account = accountname->get(src);
+				if (account && !account->empty())
```
SaberUK: The account name can never actually be empty.

You might also find it more idiomatic to write this like:

```cpp
const AccountExtItem* accountext = GetAccountExtItem();
const std::string* account = accountext ? accountext->get(src) : NULL;
if (account)
    return MOD_RES_PASSTHRU;
```

3.0/m_checkbans.cpp
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return true;
+}
+
+void CheckLists(User* source, Channel* chan, User* user)
+{
+	ListModeBase::ModeList* list;
+	ListModeBase::ModeList::const_iterator iter;
+	ModeHandler* ban = ServerInstance->Modes->FindMode("ban", MODETYPE_CHANNEL);
```
SaberUK: This + banexception should use a `ChanModeReference` now.

3.0/m_conn_require.cpp
```diff
@@ -429,7 +429,7 @@ class ModuleConnRequire : public Module
 	// This matches with 2.0's placement of OnUserInit.
 	void OnSetUserIP(LocalUser* user) CXX11_OVERRIDE
 	{
-		if (user->registered != REG_NONE)
+		if (userdata.get(user))
```
SaberUK: You unset this on `OnUserConnect` but there is no guarantee that a user's IP will not change after registration (e.g. via the IRCv3 resume spec)

PR #199, opened 2019-04-02
Port some other wanted modules to 3.0.

### LINKED ISSUE #193: m_hideidle for insp3
Hello, 

I'd like to see m_hideidle forwarded into extras for 3.0 -- it's a handy privacy feature.

Cheers,
g.

genius3000:
Porting some other wanted/requested modules to 3.0.
Please comment with any others and I'll add it to the list.

### List of modules:
- [x] m_antirandom - @SaberUK ported
- [x] m_conn_banner
- [x] m_hideidle
- [x] m_opmoderated
- [x] m_privdeaf - inspircd/inspircd#1603 to merge into m_deaf

Resolves #193.

### DIFF
--- a/3.0/m_conn_banner.cpp
+++ b/3.0/m_conn_banner.cpp
@@ -1,0 +1,67 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2012 Attila Molnar <attilamolnar@hush.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: Attila Molnar
+/// $ModAuthorMail: attilamolnar@hush.com
+/// $ModConfig: <connbanner text="Banner text goes here.">
+/// $ModDepends: core 3.0
+/// $ModDesc: Displays a static text to every connecting user before registration
+
+
+#include "inspircd.h"
+
+class ModuleConnBanner : public Module
+{
+	LocalIntExt init;
+	std::string text;
+
+ public:
+	ModuleConnBanner()
+		: init("user_init", ExtensionItem::EXT_USER, this)
+	{
+	}
+
+	void ReadConfig(ConfigStatus&) CXX11_OVERRIDE
+	{
+		text = ServerInstance->Config->ConfValue("connbanner")->getString("text");
+	}
+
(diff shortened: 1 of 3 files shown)


3.0/m_opmoderated.cpp
```diff
@@ -0,0 +1,88 @@
...
+			return MOD_RES_PASSTHRU;
+
+		Channel* const chan = target.Get<Channel>();
+		if (CheckExemption::Call(exemptionprov, user, chan, "opmoderated") == MOD_RES_ALLOW)
+			return MOD_RES_PASSTHRU;
+
+		if (!chan->GetExtBanStatus(user, 'u').check(!chan->IsModeSet(&opmod)) && chan->GetPrefixValue(user) < VOICE_VALUE)
+		{
+			MessageTarget newtarget = target;
+			newtarget.status = '@';
```
genius3000: Not sure if there's a better way to do this, without skipping the two On* calls below and sending to remote servers like timedbans does.

Robby-:
About `m_privdeaf`, maybe it makes more sense to merge this into `m_deaf` instead of keeping this as an extras module? Similar to what happened with `m_noctcp_user`, it got merged into `m_noctcp`.

3.0/m_conn_banner.cpp
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	void ReadConfig(ConfigStatus&) CXX11_OVERRIDE
+	{
+		text = ServerInstance->Config->ConfValue("connbanner")->getString("text");
+	}
+
+	void OnSetUserIP(LocalUser* user) CXX11_OVERRIDE
+	{
+		if (init.get(user))
```
SaberUK: You unset this on `OnUserConnect` but there is no guarantee that a user's IP will not change after registration (e.g. via the IRCv3 resume spec).

3.0/m_opmoderated.cpp
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	void On005Numeric(std::map<std::string, std::string>& tokens) CXX11_OVERRIDE
+	{
+		tokens["EXTBAN"].push_back('u');
+	}
+
+	Version GetVersion() CXX11_OVERRIDE
+	{
+		return Version("Implements opmoderated channel mode +U (non-voiced messages sent to ops) and extban 'u'", VF_OPTCOMMON|VF_VENDOR);
```
SaberUK:
```suggestion
		return Version("Implements opmoderated channel mode +U (non-voiced messages sent to ops) and extban 'u'", VF_OPTCOMMON);
```

3.0/m_opmoderated.cpp
SaberUK: You should be able to update the status in the existing message target.

Alternately, there's a CUList in MessageDetails that you could add all unprivileged users to.

3.0/m_opmoderated.cpp
genius3000: Will that cause a linking issue with 2.0's version and/or should I update that also?

SaberUK: It won't cause linking issues but that should be updated too.

3.0/m_conn_banner.cpp
genius3000: I'm starting to like the thought of a `OnUserPostInit` or `OnUserPreRegister` :smile:
I'd rather not leave ExtItems if I don't need to so if I check the registration state is none as well as the ExtItem it should cover all. The ExtItem will remain until connected and by then the registration state is definitely not none.

PR #236, opened 2020-04-07
Implement CI.

genius3000:
Limited to a push or pull request that changes anything in the 3.0 directory. This could maybe be improved at some point, e.g., for the next version's modules to build against that branch as well.

Installs any additional dependencies (modules should include the PackageInfo directive if necessary).

Builds are based on the inspircd/inspircd insp3 branch and only build the necessary subset of modules:
- Pull Request: build any modules within the 3.0 directory that have been changed.
- Push: build all modules within the 3.0 directory.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -1,0 +1,70 @@
+name: Linux CI
+on:
+  push:
+    paths:
+      - '3.0/*'
+  pull_request:
+    paths:
+      - '3.0/*'
+jobs:
+  build:
+    runs-on: ubuntu-16.04
+    env:
+      CXXFLAGS: -std=${{ matrix.standard }} -Wno-error=deprecated-declarations
+      CXX: ${{ matrix.compiler }}
+      INSPIRCD_DEBUG: 3
+      INSPIRCD_VERBOSE: 1
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update --assume-yes
+          sudo apt-get install --assume-yes --no-install-recommends clang g++ git make libc++-dev libc++abi-dev pkg-config
+      - name: Checkout InspIRCd@insp3
+        uses: actions/checkout@v2
+        with:
+          repository: 'inspircd/inspircd'
+          ref: 'insp3'
+          path: 'inspircd'
+      - name: Checkout self
+        uses: actions/checkout@v2
+        with:
+          path: 'contrib'
+      - name: (Pull Request) Set the build target to only the changed modules
+        if: github.event_name == 'pull_request'
+        run: |
+          cd contrib
+          git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
+          echo ::set-env name=INSPIRCD_TARGET::$( \
+          git diff --name-only origin/${{ github.base_ref }} ${{ github.sha }} | \
+          for f in `cat`; do if [[ "$f" = "3.0/"*".cpp" ]]; \
+          then echo "$(basename $f .cpp) "; fi; done)
+      - name: (Push/PR Fallback) Set the build target to all of the modules
+        if: github.event_name == 'push' || env.INSPIRCD_TARGET == ''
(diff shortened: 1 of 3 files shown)


SadieCat:
Is there a specific need to have `tools/test-build` for this? That's mainly used in inspircd/inspircd to allow us to test socket engines but modules should not be affected by this.

Instead, we can just do something like this:

```sh
make `INSPIRCD_TARGET="m_foo m_bar m_baz"
```

genius3000:
No, I was just mimicking the test of inspircd/inspircd; slimming it down here though would be good for sure.

.github/workflows/ci-macos.yml
```diff
@@ -36,12 +38,15 @@ jobs:
...
       - name: Install additional dependencies
         if: env.PACKAGES != ''
         run: brew install ${{ env.PACKAGES }}
-      - name: Run test-build-contrib
-        run: cd inspircd; ./tools/test-build-contrib c++
+      - name: Build
+        run: |
+          cd inspircd
+          ./configure --development --disable-auto-extras
+          make --jobs $(($(nproc)+1))
```
SadieCat, line 49:
```suggestion
          make --jobs $(($(getconf _NPROCESSORS_ONLN) + 1))
```

.github/workflows/ci-macos.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     runs-on: macos-latest
     env:
       CXXFLAGS: -std=${{ matrix.standard }} -Wno-error=deprecated-declarations
+      CXX: c++
```
SadieCat, line 14: There's no need to set this as its the default.

.github/workflows/ci-linux.yml
```diff
@@ -24,23 +27,25 @@ jobs:
...
       - name: Install additional dependencies
         if: env.PACKAGES != ''
         run: sudo apt-get install --assume-yes --no-install-recommends ${{ env.PACKAGES }}
-      - name: Run test-build-contrib
-        run: cd inspircd; ./tools/test-build-contrib ${{ matrix.compiler }}
+      - name: Build
+        run: |
+          cd inspircd
+          ./configure --development --disable-auto-extras
+          make --jobs $(($(nproc)+1))
```
SadieCat, line 48:
```suggestion
          make --jobs $(($(getconf _NPROCESSORS_ONLN) + 1))
```

.github/workflows/ci-macos.yml
```diff
@@ -36,12 +38,15 @@ jobs:
...
         run: |
           cd inspircd
-          echo ::set-env name=PACKAGES::$(for f in src/modules/m_*.cpp; do ./tools/directive $f PackageInfo; done)
+          echo ::set-env name=PACKAGES::$(for f in ../contrib/3.0/*; do ./tools/directive $f PackageInfo; done)
       - name: Install additional dependencies
         if: env.PACKAGES != ''
         run: brew install ${{ env.PACKAGES }}
-      - name: Run test-build-contrib
-        run: cd inspircd; ./tools/test-build-contrib c++
+      - name: Build
```
SadieCat, line 45:
```suggestion
      - name: Build the modules
```

SadieCat:
I wonder if it would be possible to get a list of changed modules from Git and only rebuild those?

genius3000:
I don't think the event supplies that info but I had seen somewhere a mention of diff'ing the whole tree by revision (or such).

I had wanted to do a full check always, but thinking about it, it would suck to have a build fail for totally unrelated reasons to one's change(s). I'll have another look at all the events we can trigger on, maybe we can trigger off the base repo or just use an interval. That way we can still do a full CI incase something in the base repo causes a breakage in an untouched module here.

SadieCat:
It's fine to do it this way for now. In the future we can look into changing it if we decide to.

SadieCat:
LGTM.

genius3000:
@SadieCat I squashed the 'build diet' and 'resolving suggestions' commits into the base 'Implement' commit. I'd appreciate another review on the commit for the PR and Push event separation. I did run it through a full PR cycle in my fork; opening a PR runs the PR check, merging a PR runs the push check.

PR #237, opened 2020-04-13
Added SAMOVE command

hoerup:


3.0/m_samove.cpp
```diff
@@ -0,0 +1,164 @@
...
+	{
+		if (parameters.size() != 3) 
+		{
+			return CMD_FAILURE;			
+		}
+
+		const std::string& nickname = parameters[0];
+		const std::string& from_channel = parameters[1];
+		const std::string& to_channel = parameters[2];
+		std::string msg = std::string();
```
SadieCat, line 59: Declaring msg here is unnecessary as you can use an empty string literal when calling PartUser.

3.0/m_samove.cpp
```diff
@@ -0,0 +1,164 @@
...
+			Channel* from_chan = ServerInstance->FindChan(from_channel);
+			if (!(from_chan))
+			{
+				user->WriteRemoteNotice("*** invalid 'from channel '" + from_channel);
+				return CMD_FAILURE;
+			}
+			Channel* to_chan = ServerInstance->FindChan(to_channel);
+			if (!(to_chan))
+			{
+				user->WriteRemoteNotice("*** invalid 'to channel '" + to_channel);
```
SadieCat, line 91:
```suggestion
				user->WriteRemoteNotice("*** invalid 'to channel' " + to_channel);
```

3.0/m_samove.cpp
```diff
@@ -0,0 +1,164 @@
...
+	CommandSamove cmd;
+ public:
+	ModuleSamove()
+		: cmd(this)
+	{
+	}
+
+	Version GetVersion() CXX11_OVERRIDE
+	{
+		return Version("Adds the /SAMOVE command which allows server operators to force move users from one channel to another.");
```
SadieCat, line 160:
```suggestion
		return Version("Adds the /SAMOVE command which allows server operators to force move users from one channel to another.", VF_OPTCOMMON);
```

3.0/m_samove.cpp
hoerup: unfortunately Channel::PartUser() doesn't accept a const string reference ( i think thats a bug?)

3.0/m_samove.cpp
hoerup: done

3.0/m_samove.cpp
hoerup: done

3.0/m_samove.cpp
SadieCat: Ah, that is a bug yeah.

For now just move the declaration next to the function cal and remove the unnecessary value (i.e. `std::string msg;`).

hoerup: Thats also what i dit in my latest commit

PR #240, opened 2020-08-28
add broadcastvendor to customtags

rgraff:
Maybe others will find this useful. We're using it in production.

Since the vendor is optional part of the tag key, this adds a config to not send the vendor.

`<customtags broadcastvendor="no">` 

Open to feedback.

### DIFF
--- a/3.0/m_customtags.cpp
+++ b/3.0/m_customtags.cpp
@@ -136,6 +136,7 @@ class CustomTags : public ClientProtocol::MessageTagProvider
 	CustomTagsExtItem ext;
 	SpecialMessageMap specialmsgs;
 	std::string vendor;
+	bool broadcastvendor;
 	int whox_index;
 
 	CustomTags(Module* mod)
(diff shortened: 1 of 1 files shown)


SadieCat:
What tags are you sending without a vendor prefix? Sending non-IRCv3 tags without a prefix is a spec violation.

rgraff:
We're sending all of our tags without a vendor prefix.  It's a closed system where we control the client and the server, so the vendor tags aren't useful to us.

I'm following the format for [IRCv3 tags](cap/message-tags):

```
<key>           ::= [ <client_prefix> ] [ <vendor> '/' ] <key_name>
```

@SadieCat which spec are you referring to it? If possible, we'd like to follow the spec.

Thanks

SISheogorath:
@rgraff Probably she is referring to:

> This means client-only tags that aren’t specified in the IRCv3 extension registry MUST use a vendor prefix and SHOULD be submitted to the IRCv3 working group for consideration if they are appropriate for more widespread adoption. See Rules for naming message tags.

As per IRCv3 message tags standard: https://ircv3.net/specs/extensions/message-tags

rgraff:
@SISheogorath Thanks. That's helpful.

In our closed system, we only have trusted tags so I didn't consider the implications on client-only tags. I see the problem with this PR and will close it.

Is it correct that the vendor prefix is optional for trusted tags and is there a way to differentiate between client and trusted when populating them? I can't find where the client_prefix is applied.

SadieCat:
Specifically, it's this part:

https://ircv3.net/specs/extensions/message-tags#standardized

>The IRCv3 Working Group reserves the right to reuse names which have not been submitted to the registry. If you do not wish to submit your tag then you MUST use a vendor-specific name (see above).

Not prefixing custom tags with a vendor-prefix means that they may conflict with IRCv3 tags in the future.

rgraff:
@SadieCat understood and thanks for the explanation

PR #244, opened 2020-11-25
Roleplay

Elizafox:
This is a 3.0 module.

It implements SCENE, SCENEA, NARRATOR, NARRATORA, AMBIANCE, FSAY, FACTION, NPC, and NPCA. SCENE, SCENEA, FACTION, and FSAY are from Charybdis (and probably from elsewhere from there); and NARRATOR, NARRATORA, NPC, NPCA, and AMBIANCE are from the old RPG module.

Quick breakdown:

SCENE and SCENEA send a message from the =Scene= user, can be used to set the scene of a roleplay session.

NARRATOR and NARRATORA send a message from the =Narrator= user, can be used for narration.

AMBIANCE sends a message from the >Ambiance< user, for compatibility with the previous module (and many other roleplay modules in other IRC daemons).

NPC and NPCA send messages from (shocker) NPC's. The nicks are underlined to avoid apparent spoofing.

FSAY and FACTION are like NPC and NPCA except that they don't underline the nicks. These were taken from Charybdis. They require operator privileges and the channels/roleplay permission.

I know this is a big one, sorry. :(

Thoroughly tested it and everything works as expected.

### DIFF
--- a/3.0/m_roleplay.cpp
+++ b/3.0/m_roleplay.cpp
@@ -672,9 +672,11 @@ public:
 	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
 	{
 		ConfigTag* tag = ServerInstance->Config->ConfValue("roleplay");
-		need_mode = tag->getBool("needchanmode", true);
 		need_op = tag->getBool("needop", false);
 		npc_host = tag->getString("npchost", "fakeuser.invalid");
+
+		// The mode can only be enabled at load-time, so check this instead
+		need_mode = (roleplaymode.GetId() != ModeParser::MODEID_MAX);
 
 		// Warn about possibly insecure configuration
 		if(!(need_mode || need_op))
--- a/3.0/m_roleplay.cpp
+++ b/3.0/m_roleplay.cpp
@@ -425,13 +425,13 @@ public:
 class CommandScene : public CommandBaseRoleplay
 {
 protected:
-	std::string GetSource(const Params& parameters)
+	std::string GetSource(const Params& parameters) CXX11_OVERRIDE
 	{
 		// Similar to the old Charybdis module
 		return "=Scene=";
 	}
 
-	std::string GetMessage(const Params& parameters)
+	std::string GetMessage(const Params& parameters) CXX11_OVERRIDE
 	{
 		return parameters[1];
 	}
@@ -447,12 +447,12 @@ public:
 class CommandSceneA : public CommandBaseRoleplay
 {
 protected:
-	std::string GetSource(const Params& parameters)
+	std::string GetSource(const Params& parameters) CXX11_OVERRIDE
 	{
 		return "=Scene=";
 	}
 
-	std::string GetMessage(const Params& parameters)
+	std::string GetMessage(const Params& parameters) CXX11_OVERRIDE
 	{
 		return MakeAction(parameters[1]);
 	}
@@ -468,13 +468,13 @@ public:
 class CommandAmbiance : public CommandBaseRoleplay
 {
 protected:
-	std::string GetSource(const Params& parameters)
+	std::string GetSource(const Params& parameters) CXX11_OVERRIDE
 	{
 		// Compatibility with the old m_rpg module
 		return ">Ambiance<";
 	}
 
-	std::string GetMessage(const Params& parameters)
+	std::string GetMessage(const Params& parameters) CXX11_OVERRIDE
 	{
 		return parameters[1];
 	}
@@ -490,12 +490,12 @@ public:
 class CommandNarrator : public CommandBaseRoleplay
 {
 protected:
-	std::string GetSource(const Params& parameters)
+	std::string GetSource(const Params& parameters) CXX11_OVERRIDE
 	{
 		return "=Narrator=";
 	}
 
-	std::string GetMessage(const Params& parameters)
+	std::string GetMessage(const Params& parameters) CXX11_OVERRIDE
 	{
 		return parameters[1];
 	}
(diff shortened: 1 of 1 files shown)


3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,704 @@
...
+			{
+				user->WriteNumeric(ERR_ROLEPLAY, c->name, InspIRCd::Format("Channel mode +%c must be set", roleplaymode.GetModeChar()));
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	inline void SendMessage(User* user, Channel* c, const std::string& source, MessageTarget& msgtarget, MessageDetails& msgdetails)
```
SadieCat, line 306: `inline` in the way it is used in this module is largely unnecessary because the functions are not exposed as part of the module's interface so the compiler can do whatever it wants with it.

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,704 @@
...
+		ConfigTag* tag = ServerInstance->Config->ConfValue("roleplay");
+		bool need_mode = tag->getBool("needchanmode", true);
+		bool need_op = tag->getBool("needop", false);
+		std::string npc_host = tag->getString("npchost", "fakeuser.invalid");
+
+		// Warn about possibly insecure configuration
+		if(!(need_mode || need_op))
+			ServerInstance->SNO->WriteToSnoMask('a', "WARNING: Roleplay configuration has needchanmode and needop both disabled, this could allow for apparent spoofing!");
+
+		// Yuck.
```
SadieCat, line 686: Maybe put these vars an anonymous namespace at the top of the module source instead?

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,704 @@
...
+		, cfaction(this, roleplaymode)
+		, cnpc(this, roleplaymode)
+		, cnpca(this, roleplaymode)
+	{
+	}
+
+	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("roleplay");
+		bool need_mode = tag->getBool("needchanmode", true);
```
SadieCat, line 678: This will not enable the mode later.

Instead of reading this in `ReadConfig` you can just do `mh->GetId() != ModeParser::MODEID_MAX` to see if the mode is registered.

SadieCat:
You may want to add the `inspircd.org/bot` to messages sent from fake clients?

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,704 @@
...
+ * Change the mode letter if needed and place the following in the channel mode
+ * list accordingly.
+
+ U                  Enable roleplay commands on this channel (requires
+                    the roleplay module)
+ */
+
+#include "inspircd.h"
+
+enum RoleplayNumerics : uint16_t
```
SadieCat, line 162: This is a C++11 extension.

```suggestion
enum RoleplayNumerics
```

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,692 @@
...
+		, cnpc(this, roleplaymode)
+		, cnpca(this, roleplaymode)
+	{
+	}
+
+	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
+	{
+		ConfigTag* tag = ServerInstance->Config->ConfValue("roleplay");
+		need_op = tag->getBool("needop", false);
+		npc_host = tag->getString("npchost", "fakeuser.invalid");
```
SadieCat, line 676: This could do with some validation.

```suggestion
		npc_host = tag->getString("npchost", "fakeuser.invalid", InspIRCd::IsHost);
```

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,692 @@
...
+		need_mode = (roleplaymode.GetId() != ModeParser::MODEID_MAX);
+
+		// Warn about possibly insecure configuration
+		if(!(need_mode || need_op))
+			ServerInstance->SNO->WriteToSnoMask('a', "WARNING: Roleplay configuration has needchanmode and needop both disabled, this could allow for apparent spoofing!");
+	}
+
+	Version GetVersion() CXX11_OVERRIDE
+	{
+		return Version("Provides commands for use in roleplay (tabletop RPG's, etc.)", VF_COMMON);
```
SadieCat, line 688:
```suggestion
		return Version("Provides commands for use in roleplay (tabletop RPGs, etc.)", VF_COMMON);
```

3.0/m_roleplay.cpp
```diff
@@ -0,0 +1,692 @@
...
+	virtual std::string GetSource(const Params&) = 0;
+
+	/* The message position can vary based on command, but this also makes a
+	 * convenient point for transforming the message (like making it an
+	 * ACTION).
+	 */
+	virtual std::string GetMessage(const Params&) = 0;
+
+	// Transform str into a CTCP action
+	inline std::string MakeAction(const std::string& str)
```
SadieCat, line 348: Same note from earlier about inline applies here.

3.0/m_roleplay.cpp
Elizafox: I think I fixed this.

CyberBotX:
So, I know this has already been merged and that the referenced issue mentions this is a rewrite and the issue that I sat on and never worked on it because of lack of time, but I would appreciate a little attribution to my version of the module that brought this module into existence. Especially with the mentions of the "old" module yet nothing mentioning my work on the old module.

Elizafox:
> So, I know this has already been merged and that the referenced issue mentions this is a rewrite and the issue that I sat on and never worked on it because of lack of time, but I would appreciate a little attribution to my version of the module that brought this module into existence. Especially with the mentions of the "old" module yet nothing mentioning my work on the old module.

I didn't mean to cause offense, sorry.

I was inspired mostly by the implementation in Charybdis; I don't know if the Charybdis module is older or not, but I certainly have known about it for many years before I found this one. I happened upon m_rpg in the old contrib modules, tried to port it, and aborted my effort as I realised I wanted to incorporate more functionality and commands, as well as cut back on code duplication. I did a total rewrite and the result doesn't bear much resemblance to your old module (even the command semantics differ a little).

I think the only ideas I took were NARRATOR, NARRATORA, the format for the AMBIANCE nick, and the idea of a setting for requiring +o or above in the channel. That, you deserve credit for.

I'm submitting a PR for this.

PR #246, opened 2020-11-28
Add m_hash_crypt module

Elizafox:
This is a crypt(3)-based hash module for compatibility with Unix crypt and other IRC daemons that use it. This relies on the system crypt which may unfortunately be deficient, hence there are ugly ifdef's.

At some point I plan on implementing shims for deficient systems, but this will do for now.

Note: I've tried to make this build on macOS, but it's a terrible idea to use this module there. macOS has a really bad crypt that doesn't support any modern hash functions. All it will support for now is DES, and needless to say this is not what you want most likely.

### DIFF
--- a/3.0/m_hash_crypt.cpp
+++ b/3.0/m_hash_crypt.cpp
@@ -170,10 +170,10 @@ class ModuleHashCrypt : public Module
 
 public:
 	ModuleHashCrypt()
-		: cryptprov_generic(this, "hash/crypt-generic", "", 2)
+		: cryptprov_generic(this, "crypt-generic", "", 2)
 #ifdef HAS_SHA
-		, cryptprov_sha256(this, "hash/crypt-sha256", "$5$", 16)
-		, cryptprov_sha512(this, "hash/crypt-sha512", "$6$", 16)
+		, cryptprov_sha256(this, "crypt-sha256", "$5$", 16)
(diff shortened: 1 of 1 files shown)


3.0/m_hash_crypt.cpp
```diff
@@ -0,0 +1,160 @@
...
+	CryptHashProvider cryptprov_blowfish;
+#endif
+#ifdef HAS_SHA
+	CryptHashProvider cryptprov_sha256;
+	CryptHashProvider cryptprov_sha512;
+#endif
+
+public:
+	ModuleHashCrypt()
+		: cryptprov_des(this, "hash/crypt-des", "", 2)  // DES is everywhere, although it's shite...
```
genius3000, line 140: This and the ones below don't require the "hash/". HashProvider will prefix that. (Please test after changing this, as I'm still unsure why things work when the name ends up as hash/hash/name).

genius3000 (commented):
Left a line comment about the names.

3.0/m_hash_crypt.cpp
Elizafox: It works

Elizafox:
I deliberately broke the macOS build because their crypt really sucks and it's not worth the trouble.

Elizafox:
I'm withdrawing this request -- I don't have the time to make this better or portable, and really it should be...

PR #247, opened 2020-11-30
Fix a bunch of issues with roleplay.

Elizafox:
**EDIT**: mention the new fixes for this.

Okay so @SadieCat gave me a good way to fix this. I took some pointers (and code) from the `m_ircv3_msgid`` module as well as her suggestions.

There are issues with the roleplay module as it stands with remote propagation: messages get duplicated, and history isn't played back correctly to clients; all roleplay commands look as if they came from the person who executed them. The reason for this is `m_spanningtree` hooks `OnUserPostMesage` to propagate messages and `m_chanhistory` uses this to store history. There are probably countless other things I missed, but those were the big two.

My first shake at this was to make a fake client class, but this could potentially cause issues as inspircd was not really designed to do this and it was a bit of a hack.

The latest commit in this branch rectifies the issues I mentioned above in another way, by using a special `inspircd.org/roleplay-src` tag that lets us rewrite things as we need to. This tag will never be seen by clients, it is simply for internal use. When we send out a user message, we now simply call `msg.SetSource()` on it to make it appear it came from us.

Previously, for posterity:
~~This is in a simlar vein to FakeUser, but inherits directly from User
since this isn't a server (although it pretends to be one, it's
ultimately just... us). This is good enough to fool most modules like
the history module and spanningtree module.

This fixes a bug where spanningtree would erroneously propagate roleplay
messages as if they came from the person who used the command. This also
happens to fix an issue where the chanhistory module would not save the
logs correctly from roleplay.~~

### DIFF
--- a/3.0/m_roleplay.cpp
+++ b/3.0/m_roleplay.cpp
@@ -180,6 +180,7 @@ namespace
 	bool need_op;
 	bool need_mode;
 	std::string npc_host;
+	std::string fakeuid;
 }
 
 /* Sigh. I had to copy this from the PRIVMSG module because it's not exported.
@@ -228,6 +229,41 @@ public:
 		// of these are restricted at the protocol level we only need to check for SOH
 		// and SPACE.
 		return (text.length() >= 2) && (text[0] == '\x1') &&  (text[1] != '\x1') && (text[1] != ' ');
+	}
+};
+
+/* We make a fake user class in the same vein as FakeUser, but not a server.
+ * We just hope nobody looks too closely at it and finds the man behind the
+ * curtain... :p
+ */
+class RoleplayUser : public User
+{
+	const std::string fake_host;
+
+public:
+	RoleplayUser(const std::string& fakenick, const std::string& fakehost)
+		: User(fakeuid, ServerInstance->FakeClient->server, USERTYPE_SERVER)
+		, fake_host(fakehost)
+	{
+		nick = fakenick;
+	}
+
+	CullResult cull() CXX11_OVERRIDE
+	{
+		// Fake users don't quit, they just get culled.
+		quitting = true;
+		// Fake users are not inserted into UserManager::clientlist or uuidlist, so we don't need to modify those here
+		return User::cull();
+	}
+
+	const std::string& GetFullHost() CXX11_OVERRIDE
+	{
+		return fake_host;
+	}
+
+	const std::string& GetFullRealHost() CXX11_OVERRIDE
+	{
+		return fake_host;
 	}
 };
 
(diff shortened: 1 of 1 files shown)


SadieCat:
This could break places that assume that a fake user can be cast to `FakeUser*` like `SplitCommand::Handle` or `IS_SERVER`.

Elizafox:
> This could break places that assume that a fake user can be cast to `FakeUser*` like `SplitCommand::Handle` or `IS_SERVER`.

As mentioned on IRC, here for posterity:

Ideas:
* Somehow (how?) construct a "hollow" `LocalUser` which acts as a message sink or something
* Add USERTYPE_UNDEFINED to `UserType`
* Set the user type to `static_cast<UserType>(0)`, which will never compare true against any other enum value... but this is definitely abuse of the type system. However, it does seem to work on GCC and clang with no warnings, and I think I've seen this idiom before.

SadieCat:
It's not ideal but for now I feel like a better option would be to send S2S as the real sender but mark it with an internal-only tag (i.e. one that always returns false for `ShouldSendTag`) and then hook `OnUserWrite` and if that tag is set then unset it and call `SetSource` to set the source.

Elizafox:
> It's not ideal but for now I feel like a better option would be to send S2S as the real sender but mark it with an internal-only (i.e. one that always returns false for `ShouldSendTag`) tag like `inspircd.org/roleplay-source=foo!bar@baz` and then hook `OnUserWrite` and if that tag is set then unset it and call `SetSource` to set the source.

This will work I think.

Elizafox:
I had to do some extra crap with copying tags to make this work properly. I looked at the `m_ircv3_msgid` module and stole the logic from there.

3.0/m_roleplay.cpp
```diff
@@ -399,22 +454,22 @@ class CommandBaseRoleplay : public Command
 	 */
 	CmdResult Handle(User* user, const Params& parameters) CXX11_OVERRIDE
 	{
-		Channel* c = ServerInstance->FindChan(parameters[0]);
 		LocalUser* luser = IS_LOCAL(user);
+		if(!luser)
+			// This shouldn't happen.
+			return CMD_FAILURE;
```
SadieCat, line 460: If the command can only executed by local users consider inheriting from SplitCommand and using HandleUser.

Elizafox: Done. :+1:

3.0/m_roleplay.cpp
```diff
@@ -344,12 +389,21 @@ class CommandBaseRoleplay : public Command
 		// Inform modules that a message is about to be sent.
 		FOREACH_MOD(OnUserMessage, (user, msgtarget, msgdetails));
 
-		ClientProtocol::Messages::Privmsg privmsg(source, c, msgdetails.text, MSG_PRIVMSG);
-		privmsg.AddTag("inspircd.org/roleplay-msg", &roleplaytag, user->nick);
+		/* Do the thing.
+		 * Tags should already have been added to msgdetails before we got here.
+		 */
+		ClientProtocol::Messages::Privmsg privmsg(user->GetFullHost(), c, msgdetails.text, MSG_PRIVMSG);
```
SadieCat, line 395: You should be able to just pass the user in directly here.

3.0/m_roleplay.cpp
```diff
@@ -427,25 +477,32 @@ class CommandBaseRoleplay : public Command
...
 		return CMD_SUCCESS;
 	}
 
 	RouteDescriptor GetRouting(User* user, const CommandBase::Params& parameters) CXX11_OVERRIDE
 	{
-		return ROUTE_OPT_BCAST;
+		/* spanningtree will broadcast this as a PRIVMSG
+		 * It would be actively harmful to broadcast this.
+		 */
+		return ROUTE_LOCALONLY;
```
SadieCat, line 505: ROUTE_LOCALONLY Is the default so you can omit GetRouting.

3.0/m_roleplay.cpp
```diff
@@ -717,6 +803,40 @@ class ModuleRoleplay : public Module
...
+		ClientProtocol::TagMap::const_iterator tag;
+
+		tag = tags.find("inspircd.org/roleplay-src");
+		if(tag == tags.end())
+			return MOD_RES_PASSTHRU;
+
+		const std::string& src = tag->second.value;
+		if(src.empty())
+			// wtf?
+			throw ModuleException("Empty roleplay source tag received");
```
SadieCat, line 829: Throwing here is potentially usafe, return MOD_RES_DENY if a message is invalid.

3.0/m_roleplay.cpp
Elizafox: Done.

3.0/m_roleplay.cpp
Elizafox: Done, and added logging.

PR #248, opened 2020-12-20
Add m_relaymsg: stateless bridging using RELAYMSG

jlu5:
[RELAYMSG](https://github.com/ircv3/ircv3-specifications/pull/417) is my take on a lightweight bridging solution for IRC. It essentially takes in /RELAYMSG commands from authorized bots and translates them as spoofed PRIVMSGs to other clients.

I've been actively using this module on my network since May (with [matterbridge](https://github.com/42wim/matterbridge/pull/1339) as the client). Hopefully this is reasonable addition here.

Note: in terms of functionality this probably has some overlap with the recent `m_roleplay`, though these are two completely independent implementations AFAIK.

### DIFF
--- a/3.0/m_relaymsg.cpp
+++ b/3.0/m_relaymsg.cpp
@@ -33,7 +33,8 @@ enum
 };
 
 // Registers the draft/relaymsg
-class RelayMsgCap : public Cap::Capability {
+class RelayMsgCap : public Cap::Capability
+{
 public:
     std::string nick_separators;
 
@@ -67,7 +68,8 @@ public:
 };
 
 // Handler for the RELAYMSG command (users and servers)
-class CommandRelayMsg : public Command {
+class CommandRelayMsg : public Command
+{
 private:
     RelayMsgCap& cap;
     RelayMsgCapTag& captag;
@@ -99,9 +101,9 @@ public:
 
     CmdResult Handle(User* user, const CommandBase::Params& parameters)
     {
-        std::string channame = parameters[0];
-        std::string nick = parameters[1];
-        std::string text = parameters[2];
+        const std::string& channame = parameters[0];
+        const std::string& nick = parameters[1];
+        const std::string& text = parameters[2];
 
         // Check that the source has the relaymsg capability.
         if (IS_LOCAL(user) && !cap.get(user)) {
--- a/3.0/m_relaymsg.cpp
+++ b/3.0/m_relaymsg.cpp
@@ -135,13 +135,7 @@ public:
         // If the sender was a lcoal user, check that the target includes a nick separator
         if (IS_LOCAL(user))
         {
-            bool accepted = false;
-            for (std::string::const_iterator x = cap.nick_separators.begin(); x != cap.nick_separators.end(); x++) {
-                if (nick.find(*x) != std::string::npos) {
-                    accepted = true;
-                }
-            }
-            if (!accepted)
+            if (nick.find_first_of(cap.nick_separators) == std::string::npos)
             {
                 user->WriteNumeric(ERR_BADRELAYNICK, nick, InspIRCd::Format("Spoofed nickname must include one of the following separators: %s", cap.nick_separators.c_str()));
                 return CMD_FAILURE;
@@ -163,7 +157,7 @@ public:
             params.push_back(nick);
             params.push_back(":" + text);
 
-            ServerInstance->PI->SendEncapsulatedData("*", "RELAYMSG", params, user);
+            ServerInstance->PI->BroadcastEncap("RELAYMSG", params, user);
         }
 
         return CMD_SUCCESS;
@@ -187,7 +181,7 @@ public:
     void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
     {
         ConfigTag* tag = ServerInstance->Config->ConfValue("relaymsg");
-        cap.nick_separators = tag->getString("separators", "/");
+        cap.nick_separators = tag->getString("separators", "/", 1);
         cmd.fake_ident = tag->getString("ident", "relay");
         cmd.fake_host = tag->getString("host", ServerInstance->Config->ServerName);
 
--- a/3.0/m_relaymsg.cpp
+++ b/3.0/m_relaymsg.cpp
@@ -1,0 +1,218 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2020 James Lu <james@overdrivenetworks.com>
+ *
+ * This file is a module for InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: jlu5
+/// $ModAuthorMail: james@overdrivenetworks.com
+/// $ModDepends: core 3
+/// $ModDesc: Provides the RELAYMSG command & draft/relaymsg capability for stateless bridging
+/// $ModConfig: <relaymsg separators="/" ident="relay" host="relay.example.com">
+//  The "host" option defaults to the local server hostname if not set.
+
+#include "inspircd.h"
+#include "modules/cap.h"
+#include "modules/ircv3.h"
+
+enum
+{
+    ERR_BADRELAYNICK = 573  // from ERR_CANNOTSENDRP in Oragono
+};
+
+// Registers the draft/relaymsg
+class RelayMsgCap : public Cap::Capability {
+public:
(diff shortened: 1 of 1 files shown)


3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "modules/cap.h"
+#include "modules/ircv3.h"
+
+enum
+{
+    ERR_BADRELAYNICK = 573  // from ERR_CANNOTSENDRP in Oragono
+};
+
+// Registers the draft/relaymsg
+class RelayMsgCap : public Cap::Capability {
```
SadieCat, line 36: You have inconsistent bracket alignment here.

```suggestion
class RelayMsgCap : public Cap::Capability
{
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    bool ShouldSendTag(LocalUser* user, const ClientProtocol::MessageTagData& tagdata) CXX11_OVERRIDE
+    {
+        return cap.get(user);
+    }
+};
+
+// Handler for the RELAYMSG command (users and servers)
+class CommandRelayMsg : public Command {
```
SadieCat, line 70: Same as above.

```suggestion
class CommandRelayMsg : public Command
{
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    std::string GetFakeHostmask(std::string& nick) {
+        return InspIRCd::Format("%s!%s@%s", nick.c_str(), fake_ident.c_str(), fake_host.c_str());
+    }
+
+    CmdResult Handle(User* user, const CommandBase::Params& parameters)
+    {
+        std::string channame = parameters[0];
+        std::string nick = parameters[1];
+        std::string text = parameters[2];
```
SadieCat, line 104: String copies are unnecessary here.

```suggestion
        const std::string& channame = parameters[0];
        const std::string& nick = parameters[1];
        const std::string& text = parameters[2];
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+        if (ServerInstance->FindNick(nick))
+        {
+            user->WriteNumeric(ERR_BADRELAYNICK, nick, "RELAYMSG spoofed nick is already in use");
+            return CMD_FAILURE;
+        }
+
+        // Make sure the nick doesn't include any core IRC characters (e.g. *, !)
+        // This should still be more flexible than regular nick checking - in particular
+        // we want to allow "/" and "~" for relayers
+        for (std::string::const_iterator x = nick.begin(); x != nick.end(); x++)
```
SadieCat, line 134: A better way to do this would be to get rid of the map and do:

```cpp
if (nick.find_first_of(""!+%@&#$:'\"?*,.";") != std::string::npos)
{
   // error
}
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+        // If the sender was a lcoal user, check that the target includes a nick separator
+        if (IS_LOCAL(user))
+        {
+            bool accepted = false;
+            for (std::string::const_iterator x = cap.nick_separators.begin(); x != cap.nick_separators.end(); x++) {
+                if (nick.find(*x) != std::string::npos) {
+                    accepted = true;
+                }
+            }
+            if (!accepted)
```
SadieCat, line 154:
```suggestion
            if (nick.find_first_of(cap.nick_separators) == std::string::npos)
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+                user->WriteNumeric(ERR_BADRELAYNICK, nick, InspIRCd::Format("Spoofed nickname must include one of the following separators: %s", cap.nick_separators.c_str()));
+                return CMD_FAILURE;
+            }
+        }
+
+        // Send the message to everyone in the channel
+        std::string fakeSource = GetFakeHostmask(nick);
+        ClientProtocol::Messages::Privmsg privmsg(fakeSource, channel, text);
+        // Tag the message as @relaymsg=<nick> so the sender can recognize it
+        privmsg.AddTag("relaymsg", &captag, user->nick);
```
SadieCat, line 165: You probably want to either draft prefix or vendor prefix this to avoid colliding with any potential ratified spec.

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+        if (IS_LOCAL(user))
+        {
+            // Pass the message on to other servers
+            CommandBase::Params params;
+            params.push_back(channame);
+            params.push_back(nick);
+            params.push_back(":" + text);
+
+            ServerInstance->PI->SendEncapsulatedData("*", "RELAYMSG", params, user);
```
SadieCat, line 176:
```suggestion
            ServerInstance->PI->BroadcastEncap("RELAYMSG", params, user);
```

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,218 @@
...
+        cap(this),
+        captag(this, cap),
+        cmd(this, cap, captag)
+    {
+    }
+
+    void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
+    {
+        ConfigTag* tag = ServerInstance->Config->ConfValue("relaymsg");
+        cap.nick_separators = tag->getString("separators", "/");
```
SadieCat, line 200: This should be ensured to be non-empty.

```suggestion
        cap.nick_separators = tag->getString("separators", "/", 1);
```

3.0/m_relaymsg.cpp
jlu5: Ooof, I totally forgot to add the prefix here too :/

It looks like the other implementations already use @relaymsg so this is a bit of a breaking change. I'll see if I can coordinate a fix?

3.0/m_relaymsg.cpp
```diff
@@ -0,0 +1,204 @@
...
+        std::string fakeSource = GetFakeHostmask(nick);
+        ClientProtocol::Messages::Privmsg privmsg(fakeSource, channel, text);
+        // Tag the message as @relaymsg=<nick> so the sender can recognize it
+        privmsg.AddTag("relaymsg", &captag, user->nick);
+        channel->Write(ServerInstance->GetRFCEvents().privmsg, privmsg);
+
+        if (IS_LOCAL(user))
+        {
+            // Pass the message on to other servers
+            CommandBase::Params params;
```
SadieCat, line 155: This won't pass along the message tags which were sent on the message. Is that intended?

If not you can probably do something like this:

```cpp
params.insert(parameters.GetTags().begin(), parameters.GetTags().end());

Comment (jlu5) [reply 3.0/m_relaymsg.cpp:155]: Hmm, I didn't really consider this case - I'm not sure if there's any use-case at the moment for a bot to send message tags in the /relaymsg call itself.

With that snippet I get this error:

```
/home/james/inspircd3/src/modules/m_relaymsg.cpp: In member function ‘virtual CmdResult CommandRelayMsg::Handle(User*, const CommandBase::Params&)’:
/home/james/inspircd3/src/modules/m_relaymsg.cpp:157:83: error: no matching function for call to ‘CommandBase::Params::insert(insp::detail::flat_map_base<std::pair<std::__cxx11::basic_string<char>, ClientProtocol::MessageTagData>, std::greater<std::__cxx11::basic_string<char> >, std::__cxx11::basic_string<char>, insp::detail::map_pair_compare<std::pair<std::__cxx11::basic_string<char>, ClientProtocol::MessageTagData>, std::greater<std::__cxx11::basic_string<char> > > >::const_iterator, insp::detail::flat_map_base<std::pair<std::__cxx11::basic_string<char>, ClientProtocol::MessageTagData>, std::greater<std::__cxx11::basic_string<char> >, std::__cxx11::basic_string<char>, insp::detail::map_pair_compare<std::pair<std::__cxx11::basic_string<char>, ClientProtocol::MessageTagData>, std::greater<std::__cxx11::basic_string<char> > > >::const_iterator)’
  157 |             params.insert(parameters.GetTags().begin(), parameters.GetTags().end());
```

Comment (jlu5) [reply 3.0/m_relaymsg.cpp:158]: (fixed below)

File: 3.0/m_relaymsg.cpp (hunk 6 above)
Comment (jlu5) [reply 3.0/m_relaymsg.cpp]: Fixed this here and in the spec https://github.com/ircv3/ircv3-specifications/pull/417/commits/70c0957bf53f5476d3b671f36ea9a4a0c086a5e

File: 3.0/m_relaymsg.cpp [hunk 10]
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
+    {
+        ConfigTag* tag = ServerInstance->Config->ConfValue("relaymsg");
+        cap.nick_separators = tag->getString("separators", "/", 1);
+        cmd.fake_ident = tag->getString("ident", "relay");
+        cmd.fake_host = tag->getString("host", ServerInstance->Config->ServerName);
+
+        if (!ServerInstance->IsIdent(cmd.fake_ident))
```
SadieCat, line 193: A possible issue: if a user edits their config and changes from a valid ident/host to an invalid ident/host then even if this throws the malformed config will sitll be applied.

To work around this you can put fake_ident and fake_host into a local variable then swap after all of the checks have been performed (also, move nick_separators to the end so it doesn't get applied till after the checks too).

PR #254, opened 2021-03-23
Added useasn configuration option

H7-25:
Added useasn configuration option to allow don't run the dns request for certain classes

### DIFF
--- a/3.0/m_asn.cpp
+++ b/3.0/m_asn.cpp
@@ -189,17 +189,8 @@ class ModuleASN CXX11_FINAL
 		if (user->quitting)
 			return;
 
-		if (user->MyClass)
-		{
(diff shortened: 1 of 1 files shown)


3.0/m_asn.cpp
```diff
@@ -189,6 +189,17 @@ class ModuleASN CXX11_FINAL
...
+		{
+			if (!user->MyClass->config->getBool("useasn", true))
+				return;
+		}
+		else
+		{
+			ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "User has no connect class in OnSetUserIP");
+			return;
+		}
+
```
SadieCat, line 202:
```suggestion
		if (!user->MyClass || !user->MyClass->config->getBool("useasn", true))
			return;
```

PR #273, opened 2023-10-26
Import m_blocksock.cpp

satmd:
Hello,

this is the module that was written by Sadie for me some months ago.


satmd:
About the module: I was getting hammered with connection requests on server-only ports and this module allowed me to silently drop those without polluting the logs.  Thank you!

PR #274, opened 2023-12-11
There's no reason extjwt module need to be a common module

H7-25:
I think there's no reason extjwt module need to be a common module.


SadieCat:
There's a HTTP endpoint which is used for validating endpoints so yeah it need to be optcommon. If you don't use that you can enable `<options:allowcommon>` to allow optcommon modules to be mismatched.

PR #276, opened 2024-07-03
add new modules for profile link and ipinfo.io in whois.

revrsedev:


4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+    {
+        CURL* curl;
+        CURLcode res;
+        std::string response;
+
+        curl_global_init(CURL_GLOBAL_DEFAULT);
+        curl = curl_easy_init();
+        if (curl)
+        {
+            std::string url = "http://ipinfo.io/" + user->client_sa.addr() + "?token=" + apikey;
```
SadieCat, line 69: This will break for UNIX socket users and potentially leaks user IPs. You probably want to add a check for `client_sa.is_ip()` and use HTTPS.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    void OnUnloadModule(Module* mod) override
+    {
+        const UserManager::LocalList& users = ServerInstance->Users.GetLocalUsers();
+        for (const auto& user : users)
+        {
+            cachedinfo.Unset(user);
+        }
+    }
```
SadieCat, line 186: Extensions are automatically unset when the providing module is unloaded. This is unnecessary.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+        else
+        {
+            new IPInfoResolver(this, IS_LOCAL(target), cachedinfo, apikey);
+        }
+    }
+
+    ModResult OnStats(Stats::Context& stats) override
+    {
+        return MOD_RES_PASSTHRU;
+    }
```
SadieCat, line 177: If you're not using this then you should not implement it.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+        {
+            cachedinfo.Unset(user);
+        }
+    }
+
+    void OnWhois(Whois::Context& whois) override
+    {
+        User* target = whois.GetTarget();
+
+        if (target->server->IsService() || target->IsModeSet('B'))
```
SadieCat, line 157: The bot mode may have its letter changed. You should use a `UserModeReference` here.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+        , Stats::EventListener(this)
+        , Whois::EventListener(this)
+        , cachedinfo(this, "ipinfo", ExtensionType::USER)
+    {
+    }
+
+    void ReadConfig(ConfigStatus& status) override
+    {
+        auto& tag = ServerInstance->Config->ConfValue("ipinfo");
+        apikey = tag->getString("apikey", "", 1);
```
SadieCat, line 144: You require `apikey` to be at least one character then but default it to an empty string anyway?

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+        curl_global_cleanup();
+    }
+
+    void ParseResponse(LocalUser* user, const std::string& response)
+    {
+        rapidjson::Document document;
+        if (document.Parse(response.c_str()).HasParseError())
+        {
+            std::lock_guard<std::mutex> lock(mtx);
+            ServerInstance->SNO.WriteGlobalSno('a', "IPInfo: Failed to parse JSON for %s: %s", user->nick.c_str(), rapidjson::GetParseError_En(document.GetParseError()));
```
SadieCat, line 94: This is not a valid format string in v4. v4 uses [fmtlib](https://fmt.dev/latest/syntax/) not snprintf.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <rapidjson/document.h>
+#include <rapidjson/error/en.h>
+#include <curl/curl.h>
+#include <thread>
+#include <mutex>
+
+enum
+{
+    // deff custom numeric for this information in whois
+    RPL_WHOISIPINFO = 695,
```
SadieCat, line 41: You may find it better to use `RPL_WHOISSPECIAL` (320) here as it is more of a generic numeric for text strings relating to a user that doesn't need special client support.

4/m_profileLink.cpp
```diff
@@ -0,0 +1,73 @@
...
+		// Skip services, bots.
+		if (target->server->IsService() || target->IsModeSet('B'))
+			return;
+
+		// Check if the user has an account name (is registered).
+		const std::string* account = accountapi->GetAccountName(target);
+		if (account)
+		{
+			// Construct the profile URL using the user's account name.
+			const std::string profileUrl = "https://www.reseau-entrenous.fr/profil/" + *account;
```
SadieCat, line 61: Is this a network specific thing? We only accept modules into contrib that are usable on multiple networks.

4/m_profileLink.cpp
```diff
@@ -0,0 +1,73 @@
...
+	void OnWhois(Whois::Context& whois) override
+	{
+		User* target = whois.GetTarget();
+
+		// Skip services, bots.
+		if (target->server->IsService() || target->IsModeSet('B'))
+			return;
+
+		// Check if the user has an account name (is registered).
+		const std::string* account = accountapi->GetAccountName(target);
```
SadieCat, line 57: This will result in a null dereference if the account module is not loaded. You may want to add a check for that beforehand.

4/m_profileLink.cpp
```diff
@@ -0,0 +1,73 @@
...
+		, accountapi(this)
+	{
+	}
+
+	void OnWhois(Whois::Context& whois) override
+	{
+		User* target = whois.GetTarget();
+
+		// Skip services, bots.
+		if (target->server->IsService() || target->IsModeSet('B'))
```
SadieCat, line 53: The bot mode may have its letter changed. You should use a `UserModeReference` here.

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        const std::string* cached = cachedinfo.Get(target);
+        if (cached)
+        {
+            whois.SendLine(RPL_WHOISSPECIAL, "ip info(cached): " + *cached);
+        }
+        else
+        {
+            new IPInfoResolver(this, IS_LOCAL(target), cachedinfo, apikey);
```
SadieCat, line 198: If you have multiple servers this `IS_LOCAL` will result in a null dereference in the constructor.

revrsedev: oh boy links thanks

4/m_ipinfo_io.cpp
```diff
@@ -0,0 +1,213 @@
...
+            curl_global_cleanup();
+        }
+
+        void ParseResponse(User* user, const std::string& response)
+        {
+            rapidjson::Document document;
+            if (document.Parse(response.c_str()).HasParseError())
+            {
+                std::lock_guard<std::mutex> lock(mtx);
+                ServerInstance->SNO.WriteGlobalSno('a', fmt::format("IPInfo: Failed to parse JSON for {}: {}", user->nick, rapidjson::GetParseError_En(document.GetParseError())));
```
SadieCat, line 89: The method already takes a fmtlib format string there's no need to call fmt::format here.

PR #281, opened 2025-01-07
add module GeoLite Citys.

revrsedev:
add module GeoLite Citys to whois response.


SadieCat:
Please only submit changes one module per pull request and make sure the Git history is clean so I can actually see what has changed.

4/m_censor.cpp
```diff
@@ -42,22 +36,166 @@ class ModuleCensor : public Module
...
+			return false;
+		}
+
+		return whitelist_matcher->matches(status) || IsEmojiOnly(text) || IsKiwiIRCOnly(text);
+	}
 
  public:
 	ModuleCensor()
-		: Module(VF_NONE, "Allows the server administrator to define inappropriate phrases that are not allowed to be used in private or channel messages.")
+		: Module(VF_NONE, "Allows the server administrator to define inappropriate phrases that are not allowed to be used in private or channel messages and blocks messages with mixed UTF-8 scripts, only allowing certain Unicode smileys.")
```
SadieCat, line 142: This should probably just be a new module rather than tacking it on to an existing one (which needs to keep compatibility with the v3 version).

4/m_whoisgeolite.cpp
```diff
@@ -0,0 +1,139 @@
...
+	void ReadConfig(ConfigStatus& status) override
+	{
+		// Load configuration and get the path to the GeoLite2 database
+		auto& tag = ServerInstance->Config->ConfValue("geolite");
+		dbpath = tag->getString("dbpath", "/etc/GeoLite2-City.mmdb");
+
+		// Attempt to open the MaxMind GeoLite2-City database
+		int status_open = MMDB_open(dbpath.c_str(), MMDB_MODE_MMAP, &mmdb);
+		if (status_open != MMDB_SUCCESS) {
+			ServerInstance->SNO.WriteGlobalSno('a', "GeoLite2: Failed to open GeoLite2 database: " + std::string(MMDB_strerror(status_open)));
```
SadieCat, line 62: You probably just want to throw a ModuleException here because it doesn't make sense to have the module loaded without being configured. That will allow you to get rid of `db_loaded`.

4/m_whoisgeolite.cpp
```diff
@@ -0,0 +1,139 @@
...
+		User* source = whois.GetSource();  // The user issuing the WHOIS command
+		User* target = whois.GetTarget();  // The user being WHOIS'd
+
+		// Only allow IRC operators to see the city information
+		if (!source->IsOper()) {
+			return;
+		}
+
+		// Check if the user is remote
+		if (!IS_LOCAL(target)) {
```
SadieCat, line 81: You check IS_LOCAL twice here which doesn't make sense as a user won't suddenly become non-local.

SadieCat: Also if you look up when the remote address changes and then store it in a synced StringExtItem it will allow remote users to see the country users are connecting from.

4/m_whoisgeolite.cpp
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		// Perform the GeoLite2 lookup using the socket address
+		int gai_error = 0;
+		const struct sockaddr* addr = reinterpret_cast<const struct sockaddr*>(&luser->client_sa);
+		MMDB_lookup_result_s result = MMDB_lookup_sockaddr(&mmdb, addr, &gai_error);
+
+		if (gai_error != 0) {
+			ServerInstance->SNO.WriteGlobalSno('a', "GeoLite2: getaddrinfo error: " + std::string(gai_strerror(gai_error)));
+			whois.SendLine(RPL_WHOISSPECIAL, "*", "City: Unknown (lookup error).");
```
SadieCat, line 107: `RPL_WHOISSPECIAL` doesn't take two parameters, get rid of the `"*"`.

SadieCat:
You probably want to give the module a more descriptive name, e.g. `m_whoisgeocity`.

4/m_whoisgeolite.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+		// Ensure the MaxMind database is loaded
+		if (!db_loaded) {
+			whois.SendLine(RPL_WHOISSPECIAL, "*", "City: GeoLite2 database not loaded.");
+			return;
+		}
+
+		// Perform the GeoLite2 lookup using the socket address
+		int gai_error = 0;
+		const struct sockaddr* addr = reinterpret_cast<const struct sockaddr*>(&luser->client_sa);
```
SadieCat, line 102: You can just use `&sa.sa` instead of casting this.

4/m_whoisgeolite.cpp
```diff
@@ -18,130 +18,90 @@
...
 
-/// $CompilerFlags: find_compiler_flags("libmaxminddb")
-/// $LinkerFlags: find_linker_flags("libmaxminddb")
-
-/// $PackageInfo: require_system("alpine") libmaxminddb-dev pkgconf
-/// $PackageInfo: require_system("arch") pkgconf libmaxminddb
-/// $PackageInfo: require_system("darwin") libmaxminddb pkg-config
-/// $PackageInfo: require_system("debian~") libmaxminddb-dev pkg-config
-/// $PackageInfo: require_system("rhel~") pkg-config libmaxminddb-devel
+/// $LinkerFlags: -lmaxminddb
```
SadieCat: Removing these directives will break the CI build and building on distros which don't have a global library path.

PR #282, opened 2025-01-09
add new module Google reCAPTCHA

revrsedev:
This module integrates Google reCAPTCHA v2 "Im not a robot" with InspIRCd-4 to protect against automated bots. It requires users to solve a CAPTCHA before they can join certain channels to perform specific actions. Nickserv accounts and whitelisted channels can bypass.


revrsedev:
Closed for now to many things that i need to re-do

PR #283, opened 2025-01-10
add new module Google recaptcha V2

revrsefr:
Google recaptcha to IRCd's.

### DIFF
--- a/4/m_recaptchav2.cpp
+++ b/4/m_recaptchav2.cpp
@@ -8,24 +8,25 @@
  * redistribute it and/or modify it under the terms of the GNU General Public
  * License as published by the Free Software Foundation; version 2.
  */
-​
+
 /// $ModAuthor: reverse mike.chevronnet@gmail.com
-/// $ModConfig: <captchaconfig conninfo="dbname=example user=postgres password=secret hostaddr=127.0.0.1 port=5432" mode="master/slave" masterserver="master.example.com" url="http://meme.com/verify/" whitelistchan="#help,#support" whitelistport="6697,6666">
+/// $ModConfig: <captchaconfig conninfo="dbname=example user=postgres password=secret hostaddr=127.0.0.1 port=5432" 
+/// mode="master/slave" masterserver="master.example.com" url="http://meme.com/verify/" whitelistchan="#help,#support">
 /// $ModDepends: core 4
-​
+
 /// $CompilerFlags: find_compiler_flags("libpq")
 /// $LinkerFlags: find_linker_flags("libpq")
-​
+
 #include "inspircd.h"
 #include "extension.h"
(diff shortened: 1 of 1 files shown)


4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,412 @@
...
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; version 2.
+ */
+
+/// $ModAuthor: reverse mike.chevronnet@gmail.com
+/// $ModConfig: <captchaconfig conninfo="dbname=example user=postgres password=secret hostaddr=127.0.0.1 port=5432" 
+/// mode="master/slave" masterserver="master.example.com" url="http://meme.com/verify/" whitelistchan="#help,#support">
+/// $ModDepends: core 4
+
+/// $CompilerFlags: find_compiler_flags("libpq")
```
SadieCat, line 17: It might be better if you use the InspIRCd SQL API instead of directly linking against postgres as its designed to integrate into the mainloop and not block connections if your SQL server is slow.

4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,412 @@
...
+        if (count > 0)
+        {
+            ip_cache[ip] = now + std::chrono::minutes(CACHE_DURATION_MINUTES);
+            return true;
+        }
+
+        return false;
+    }
+
+    void OnEncapsulatedData(const std::string& source, const std::string& command, CommandBase::Params& parameters)
```
SadieCat, line 328: OnEncapsulatedData???

4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+            response_params.push_back("check");
+            response_params.push_back(ip);
+            response_params.push_back((PQresultStatus(res) == PGRES_TUPLES_OK && atoi(PQgetvalue(res, 0, 0)) > 0) ? "success" : "failure");
+            ServerInstance->PI->SendEncapsulatedData(source, "RECAPTCHA-REPLY", response_params);
+
+            PQclear(res);
+        }
+    }
+    else if (command == "RECAPTCHA-REPLY")
```
SadieCat, line 382: given this just snotices it would be simpler to just use WriteGlobalSno.

4/m_recaptchav2.cpp
revrsefr: Yeah, it's a first for me server's comunications, im a little confuse how to handle encapsulated commands

SadieCat: You just handle them as normal commands all encapsulation does is quietly fail if the command isn't available instead of causing a delink.

PR #284, opened 2025-01-11
add new module Google reCAPTCHA V2

revrsefr:


4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        void OnError(const SQL::Error& error) override
+        {
+            ServerInstance->SNO.WriteGlobalSno('r', INSP_FORMAT("*** reCAPTCHA: SQL Error: {}", error.ToString()));
+            user->WriteNotice("*** reCAPTCHA: Failed to add IP to the whitelist.");
+        }
+    };
+
+    class CheckIPQuery : public SQL::Query
```
SadieCat, line 78: Broken indentation.

4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,292 @@
...
+    CommandRecaptcha cmd;
+    Account::API accountapi;
+
+    friend class CommandRecaptcha;
+
+    class AddIPQuery : public SQL::Query
+    {
+    private:
+        ModuleCaptchaCheck* parent;
+        User* user;
```
SadieCat, line 59: If the user disconnects before the query completes then this will result in a use after free. You should store the UUID In the class and then call `ServerInstance->Users.FindUUID` to look up the user.

4/m_recaptchav2.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+            return MOD_RES_DENY;
+        }
+
+        return MOD_RES_PASSTHRU;
+    }
+
+    void OnUserPostInit(LocalUser* user) override
+    {
+        const std::string* status = captcha_success.Get(user);
```
SadieCat, line 232: This isn't actually set anywhere? Also, you could use BoolExtItem here to reduce memory.

PR #285, opened 2025-01-25
add m_geomaxlite and deleted m_whoisgeolite.

revrsefr:
Enhance: Add and usermode +y and rename the module.


SadieCat:
`4/\342\200\216m_geomaxlite.cpp`??? I think you added some unintentional characters there.

4/‎m_geomaxlite.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: reverse <mike.chevronnet@gmail.com>
+/// $ModDesc: Adds city and country information to WHOIS using local MaxMind database and with it's own usermode +y.
+/// $ModConfig: <geolite dbpath="path/geodata/GeoLite2-City.mmdb">
+/// $ModDepends: core 4
+
+/// $LinkerFlags: -lmaxminddb
```
SadieCat, line 24: You were using pkg-config before but now you're not, this will break building on systems where libmaxminddb.so isn't in the default library path.

4/‎m_geomaxlite.cpp
```diff
@@ -0,0 +1,130 @@
...
+/// $PackageInfo: require_system("darwin") libmaxminddb pkg-config
+/// $PackageInfo: require_system("debian~") libmaxminddb-dev pkg-config
+/// $PackageInfo: require_system("rhel~") pkg-config libmaxminddb-devel
+
+#include "inspircd.h"
+#include "modules/whois.h"
+#include "extension.h"
+#include <maxminddb.h>
+
+class GeoLiteMode final : public SimpleUserMode
```
SadieCat, line 37: If you're not going to customise the mode behaviour there's no need to subclass it just use `SimpleUserMode` directly.

4/‎m_geomaxlite.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+        std::string city = (status_city == MMDB_SUCCESS && city_data.has_data) ? std::string(city_data.utf8_string, city_data.data_size) : "Unknown";
+        std::string country = (status_country == MMDB_SUCCESS && country_data.has_data) ? std::string(country_data.utf8_string, country_data.data_size) : "Unknown";
+
+        country_item.Set(user, "City: " + city + ", Country: " + country);
+    }
+
+    void OnUserQuit(User* user, const std::string& message, const std::string& opermessage) override
+    {
+        if (user->IsModeSet(geolite_mode))
```
SadieCat, line 120: You don't need to check if the mode is set before unsetting here as unsetting an extension on a user who doesn't have it set is a no-op.

4/‎m_geomaxlite.cpp
```diff
@@ -0,0 +1,130 @@
...
+            country_item.Unset(user);
+    }
```
SadieCat, line 122: This is unnecessary as extensions are automatically unset on quitting users.

4/‎m_geomaxlite.cpp
revrsefr: yes already add then again

PR #292, opened 2025-03-14
add IRCV3 +draft/typing

revrsefr:


SadieCat:
I'm not sure about the purpose of this? You can already use typing notifications with InspIRCd since v3? It doesn't require a special capability it just comes under `message-tags`.

revrsefr:
ohh make no sense ill close

PR #293, opened 2025-03-16
add module Google recaptcha with JWT tokens.

revrsefr:
New module Google recaptcha with JWT tokens.


SadieCat:
You need to provide PackageInfo comments for at the least Debian/Ubuntu to allow the CI to check the module builds.

PR #300, opened 2026-01-14
Add InspIRCd 4 security-groups module

revrsefr:
Title
Add InspIRCd 4 security-groups module

Description
This PR adds m_securitygroups.cpp, an InspIRCd 4 module implementing security-groups.

Features

Defines <securitygroup ...> config blocks with match criteria:
mask (hostmask globs and CIDR ranges)
account/registered
tls/tls-users and insecure/insecure-users
websocket/websocket-users
webirc/webirc-users
scoremin / scoremax (via the m_reputation module user ext, if present) 
Exposes user extensions:
securitygroups (all matched groups; synced network-wide)
securitygroups-public (public subset; synced network-wide)
Adds /SECURITYGROUPS [nick] (shows full list to self/auspex, otherwise public list)
Adds extban g: / sg to match users by security-group name (supports glob matching)
Notes

If m_account is not loaded, account/registered requirements will not match.
If the reputation user extension is not available, scoremin/scoremax requirements will not match.
Testing

### DIFF
--- a/4/m_securitygroups.cpp
+++ b/4/m_securitygroups.cpp
@@ -1,0 +1,381 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   (C) 2026 reverse - mike.chevronnet@gmail.com
+ *
+ * Implements UnrealIRCd 6-style security-groups for InspIRCd 4.
+ *
+ * Configuration example:
+ *
+ *   <securitygroup name="webirc-users"
+ *     mask="*@webirc.example.com"
+ *     webirc="yes"
+ *     scoremin="0"
+ *     scoremax="100"
+ *     public="yes">
+ *
+ * Supported matching options:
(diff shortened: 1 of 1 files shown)


4/m_securitygroups.cpp
```diff
@@ -0,0 +1,381 @@
...
+				return false;
+		}
+
+		if (group.require_webirc && !webircext.Get(user))
+			return false;
+
+		if (group.require_websocket && !IsWebSocketUser(user))
+			return false;
+
+		const bool istls = (SSLIOHook::IsSSL(&user->eh) != nullptr);
```
sadiepowell: You should use IsSecure or GetCertificate from the SSL API here because a user might be connecting securely using a proxy over UNIX sockets or connecting insecurely via a proxy that is connected using TLS.

sadiepowell (changes requested):

PR #303, opened 2026-06-07
Fix priority for /swhois ADD

Kufat:
```
ADD <nick> [<priority>] :<message>
^0  ^1     ^2            ^back
```


sadiepowell:
Good catch!

PR #304, opened 2026-06-07
m_swhois_ext: Fix flag processing and sorting by priority

Kufat:
`== npos` was being used instead of `!= npos` so we only got the flags we _didn't_ want.

Priority was being set after the swhois list had been sorted, so it wasn't doing any good. I thought about making the priority an optional arg and defaulting it to zero, but it felt like it'd be too easy to forget unintentionally. If you'd rather do it that way I can change it, though.


sadiepowell:

PR #305, opened 2026-06-16
Add m_ircv3_preaway

revrsefr:
> Adds m_ircv3_preaway, implementing draft/pre-away (https://ircv3.net/specs/extensions/pre-away). CommandAway already sets works_before_reg, so the server accepts AWAY during registration as-is — this just advertises the cap. Tested: cap shows in CAP LS; AWAY :x before CAP END returns 306; post-registration WHOIS shows 301 :x. Draft, happy to iterate.

### DIFF
--- a/4/m_ircv3_preaway.cpp
+++ b/4/m_ircv3_preaway.cpp
@@ -1,0 +1,43 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2026 reverse <mike.chevronnet@gmail.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: reverse <mike.chevronnet@gmail.com>
+/// $ModDepends: core 4
+/// $ModDesc: Provides the DRAFT pre-away IRCv3 extension.
+
+
+#include "inspircd.h"
+#include "modules/cap.h"
+
+class ModuleIRCv3PreAway final
+	: public Module
+{
+private:
+	Cap::Capability cap;
+
+public:
+	ModuleIRCv3PreAway()
+		: Module(VF_VENDOR, "Provides the IRCv3 draft/pre-away client capability.")
+		, cap(this, "draft/pre-away")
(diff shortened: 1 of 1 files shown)


sadiepowell:
Pre-away also has some special casing regarding `*` does it not?

4/m_ircv3_preaway.cpp
```diff
@@ -0,0 +1,74 @@
...
+	{
+		// draft/pre-away gives "*" a special meaning: the user is away but isn't
+		// saying why. The spec lets the server swap in a human-readable string before
+		// that gets relayed to other clients, which is what we do here.
+		//
+		// The spec note about "*" not overriding an away message set by another of the
+		// user's connections doesn't apply to us: InspIRCd tracks away state per
+		// connection, so there's no shared session to clobber.
+		if (!substitute.empty() && message == "*")
+			message = substitute;
```
sadiepowell: I've asked for spec clarification in the IRCv3 channel but I don't think this is correct. I think the asterisk is supposed to be substituted for clients without the cap but not clients with the cap so you probably want to intercept `RPL_AWAY` and away-notify `AWAY` messages with a capability check.

4/m_ircv3_preaway.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+		if (command == "AWAY")
+		{
+			User* source = msg.GetSourceUser();
+			if (!source)
+				return MOD_RES_PASSTHRU;
+
+			ClientProtocol::Message rewritten("AWAY", source);
+			rewritten.PushParam(substitute);
+			user->Send(awayprov, rewritten);
```
sadiepowell: I haven't checked this but doesn't this cause a loop? The sent message will also call OnUserWrite.

4/m_ircv3_preaway.cpp
```diff
@@ -0,0 +1,89 @@
...
+	: public Module
+{
+private:
+	Cap::Capability cap;
+	ClientProtocol::EventProvider awayprov;
+	std::string substitute;
+
+public:
+	ModuleIRCv3PreAway()
+		: Module(VF_VENDOR, "Provides the DRAFT draft/pre-away IRCv3 client capability.")
```
sadiepowell:
```suggestion
		: Module(VF_NONE, "Provides the DRAFT draft/pre-away IRCv3 client capability.")
```

4/m_ircv3_preaway.cpp
```diff
@@ -0,0 +1,89 @@
...
+		, cap(this, "draft/pre-away")
+		, awayprov(this, "AWAY")
+	{
+	}
+
+	void ReadConfig(ConfigStatus&) override
+	{
+		const std::string sub = ServerInstance->Config->ConfValue("preaway")->getString("substitute", "Away");
+		substitute = (sub == "*") ? "" : sub;
```
sadiepowell:
```suggestion
		substitute = ServerInstance->Config->ConfValue("preaway")->getString("substitute", "Away", [](const auto& str) {
			return !str.empty() && str.length() <= ServerInstance->Config->Limits.MaxAway && !irc::equals(str, "*");
		});
```

Untested but you should probably do something like this because you can't have an empty away message and it removes the need to check for it later.

PR #308, opened 2026-06-16
Add m_ircv3_message_redaction.

revrsefr:
Implements the message-redaction draft: https://ircv3.net/specs/extensions/message-redaction

Adds a REDACT command + draft/message-redaction cap. A client deletes one of its own messages by msgid (REDACT []) and the server relays the REDACT to recipients who have the cap so their clients can hide/log it. Channel mods (rank ≥ minrank) and opers with channels/redact can redact others' messages; self-redaction can be time-limited with .

To authorise it I track recent msgid → {sender, target, time}, pruned by and capped by . Works across a link since msgids and uuids are network-global.

### DIFF
--- a/4/m_ircv3_message_redaction.cpp
+++ b/4/m_ircv3_message_redaction.cpp
@@ -1,0 +1,309 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2026 reverse <mike.chevronnet@gmail.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $ModAuthor: reverse <mike.chevronnet@gmail.com>
+/// $ModConfig: <redaction window="0" retention="1h" minrank="20000" operoverride="yes" maxtracked="200000">
+/// $ModDepends: core 4
+/// $ModDesc: Provides the DRAFT IRCv3 draft/message-redaction extension (REDACT command).
+
+// draft/message-redaction lets a client delete one of its previously sent messages
+// by msgid: REDACT <target> <msgid> [<reason>]. Once the redactor is authorised the
+// server relays ":<source> REDACT <target> <msgid> [<reason>]" to recipients that
+// negotiated the cap, so their clients can hide or log the deletion.
+//
+// Who may redact (all under <redaction>):
+//   - the original sender (matched by uuid or services account), within <window>
+//     seconds of sending it (0 = no limit);
+//   - a channel moderator with rank >= minrank (default halfop);
+//   - an oper holding channels/redact, when operoverride is enabled.
+//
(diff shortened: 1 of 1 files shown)


4/m_ircv3_message_redaction.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+	// config
+	time_t window = 0;                 // self-redaction time limit (0 = unlimited)
+	time_t retention = 3600;           // how long a msgid stays redactable
+	unsigned long minrank = HALFOP_VALUE; // min channel rank for moderator redaction
+	bool operoverride = true;          // opers with channels/redact can redact anything
+	size_t maxtracked = 200000;        // hard cap on the tracking map
+
+	ModuleRedaction()
+		: Module(VF_VENDOR | VF_COMMON, "Provides the DRAFT IRCv3 draft/message-redaction extension (REDACT command).")
```
sadiepowell: This isn't a vendor module.


```suggestion
		: Module(VF_COMMON, "Provides the DRAFT IRCv3 draft/message-redaction extension (REDACT command).")
```

4/m_ircv3_message_redaction.cpp
```diff
@@ -0,0 +1,309 @@
...
+		, cmd(this, *this)
+	{
+	}
+
+	void ReadConfig(ConfigStatus&) override
+	{
+		const auto& tag = ServerInstance->Config->ConfValue("redaction");
+		window = tag->getDuration("window", 0, 0);
+		retention = tag->getDuration("retention", 3600, 60);
+		minrank = tag->getNum<unsigned long>("minrank", HALFOP_VALUE, 0);
```
sadiepowell: This should be ModeHandler::Rank in case the rank type changes.

4/m_ircv3_message_redaction.cpp
```diff
@@ -0,0 +1,309 @@
...
+	penalty = 2000;
+}
+
+RouteDescriptor CommandRedact::GetRouting(User*, const Params& parameters)
+{
+	const std::string& t = parameters[0];
+	if (!t.empty() && ServerInstance->Channels.IsPrefix(t[0]))
+		return ROUTE_BROADCAST; // every server delivers to its local channel members
+	if (User* tu = ServerInstance->Users.FindNick(t, true))
+		return ROUTE_OPT_UCAST(tu->server); // only the target's server needs it
```
sadiepowell: Its unusual to use optional unicast for one but broadcast for the other, did you mean to make both optional or neither? Given the module is VF_COMMON it can probably just be neither.

4/m_ircv3_message_redaction.cpp
```diff
@@ -0,0 +1,309 @@
...
+	// Remember a freshly sent message keyed by its msgid (set by m_ircv3_msgid).
+	void Track(User* user, const MessageTarget& target, const ClientProtocol::TagMap& tags)
+	{
+		std::string tname;
+		if (target.type == MessageTarget::TYPE_CHANNEL)
+			tname = target.Get<Channel>()->name;
+		else if (target.type == MessageTarget::TYPE_USER)
+			tname = target.Get<User>()->nick;
+		else
+			return; // not a redactable target
```
sadiepowell: `target.GetName()` probably does what you want here.

4/m_ircv3_message_redaction.cpp
```diff
@@ -0,0 +1,309 @@
...
+		window = tag->getDuration("window", 0, 0);
+		retention = tag->getDuration("retention", 3600, 60);
+		minrank = tag->getNum<unsigned long>("minrank", HALFOP_VALUE, 0);
+		operoverride = tag->getBool("operoverride", true);
+		maxtracked = tag->getNum<size_t>("maxtracked", 200000, 1000);
+	}
+
+	std::string AccountOf(User* user) const
+	{
+		const std::string* acct = accountapi ? accountapi->GetAccountName(user) : nullptr;
```
sadiepowell: It might be safer to use account identifiers rather than names here?

PR #311, opened 2026-07-01
Update m_ircv3_extended_isupport to use the core ISupport API

revrsefr:
Now that the ISupport API is in core, the module can send the list with SendTo() instead of rebuilding the 005 tokens itself. Batching moved to OnPreEventSend/OnPostEventSend so the whole burst goes out as one draft/extended-isupport batch, and it uses OnSendISupportDiff for early requests.

Depends on the new ISupport::API / WriteNumeric(vector) / GetNumeric / OnPostEventSend commits (on insp4).

### DIFF
--- a/4/m_ircv3_extended_isupport.cpp
+++ b/4/m_ircv3_extended_isupport.cpp
@@ -22,6 +22,176 @@
 
 #include "inspircd.h"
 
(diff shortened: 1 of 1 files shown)


revrsefr:
Hi,
Im actually writing/running my own ircd and will soon post it on Github. I'm no longer doing anything for inspircd at this point.

